Issue #5, opened 2023-01-31
Add optional caching layer


willscott:
this may be a separate project / component

willscott:
see #9  - I'll move this to bifrost-gateway once https://github.com/ipfs/github-mgmt/pull/113 occurs

aarshkshah1992:
Based on the discussion, I assume there's no work to be done here for the Caboose folks then ?

willscott:
correct

PR #9, opened 2023-02-06
Proposal for simple request-level block cache

willscott:
Maybe we do an A-B test to see what's more effective between a single LRU cache across all accesses, versus this model of keeping a small cache that's only scoped to inbound HTTP requests.

I can make libraries for both and we can come up with a config flag for which one(s) we turn on.

### DIFF
--- a/cache/context_cache.go
+++ b/cache/context_cache.go
@@ -1,0 +1,155 @@
+package cache
+
+import (
+	"context"
+	"errors"
+	"sync"
+
+	lru "github.com/hashicorp/golang-lru"
+	"github.com/ipfs/go-cid"
+	ipfsblockstore "github.com/ipfs/go-ipfs-blockstore"
+	"github.com/ipfs/go-libipfs/blocks"
+)
+
+type ContextCache struct {
+	backing ipfsblockstore.Blockstore
+
+	ctxkey string
+	cache  sync.Map
+}
+
+var (
+	ErrNotImplemented error = errors.New("not implemented")
+	ErrInternalError  error = errors.New("internal error")
+)
+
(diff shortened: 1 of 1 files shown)


willscott:
fix #5

willscott:
@aschmahmann may have thoughts on the right caching things to prototype / experiment with in this context

aschmahmann:
@willscott yeah, I agree that we're going to need some experimentation here and that the two thoughts you mentioned were going to be the easiest to start with. However, there's a wide range of options here that we may have to explore and A-B test to figure out what's best.

However, IMO these components should probably live in bifrost-gateway rather than caboose as it minimizes the scope of the component to basically do the saturn-specific things and makes it easier for them to own it. What we end up going with ends up being pretty specific to that deployment rather than the Saturn logic of where to get stuff.

If the components end up being non-trivial and reusable they could end up moving to go-libipfs as well.

aschmahmann:
Here's one that uses a global LRU in bifrost-gateway https://github.com/ipfs/bifrost-gateway/pull/15.

I'd like us to have the option of a request-level cache as well, probably leveraged by using sessions on the exchange interface. I'm trying to avoid calls like `Has` living in Caboose though as they'll likely end up breaking abstraction layers (i.e. we do not have something if it maybe lives in a remote node multiple hops away).

aarshkshah1992:
IIUC, this will now be moved to the Gateway, right ?

lidel:
@aarshkshah1992 we have global block cache based on [TwoQueueCache](https://pkg.go.dev/github.com/hashicorp/golang-lru/v2#TwoQueueCache):  https://github.com/ipfs/bifrost-gateway/blob/main/blockstore_cache.go

Request-level could be added,  if we find the need for it, but  this should happen in biforst-gateway (which has understanding of requested content paths), and not caboose.

PR #12, opened 2023-02-07
retry failed requests

### LINKED ISSUE #3: fail-over data retrieval
currently if the first request fails an error is propagated upwards. instead, the library should attempt a fail-over backend as well

willscott:
fix #3

### DIFF
--- a/pool.go
+++ b/pool.go
@@ -150,18 +150,33 @@ func (p *pool) Close() {
 	}
 }
 
-func (p *pool) fetchWith(ctx context.Context, c cid.Cid, with string) (blocks.Block, error) {
+func (p *pool) fetchWith(ctx context.Context, c cid.Cid, with string) (blk blocks.Block, err error) {
 	<-p.started
-	aff := with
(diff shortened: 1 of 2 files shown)


pool.go
```diff
@@ -153,18 +158,33 @@ func (p *pool) Close() {
...
+		aff := with
+		if aff == "" {
+			aff = fmt.Sprintf("%d%s", i, c.Hash().B58String())
+		} else {
+			aff = fmt.Sprintf("%d%s", i, aff)
+		}
+		p.lk.RLock()
+		member := p.c.LocateKey([]byte(aff))
+		p.lk.RUnlock()
+		root := member.String()
```
hacdias, line 179: Can't we keep this outside the loop? Probably netgligable computational complexity, but still.

willscott: we want to fail-over to a different backend when the first one fails - so we re-locate a different member by putting the different retry number into the key we locate

hacdias (approved):

Issue #16, opened 2023-02-08
error: lack of robust retry on multiple error responses from specific PoPs


lidel:
Sometimes, Saturn PoPs return 500. See https://github.com/ipfs/bifrost-gateway/issues/16#issuecomment-1422104572

## Proposed improvement

- If this happens more than once, we should ignore that PoP and retry with a different one
- If multiple PoPs fail, we run out of PoPs, we are unable to serve requests.
  - We need a mechanism where we can ask orchestrator to exclude faulty ones, e.g., `https://orchestrator.strn.pl/nodes/nearby?exclude=IP1,IP2,IP3`

PR #19, opened 2023-02-09
Use a more advanced weighting / fail-over strategy

### LINKED ISSUE #1: Dynamic Load
currently a stable hash is used that weights all current backends equally.

the relative weights of backends should adapt based on observed performance.

willscott:
fix #1 

this is a second pass at the consistent hash

* this adds logic to decrease the weight of provided backends each time they fail a request.

As-is this should make the prevalence of 503's quite a bit better.

Todo before done:
* [x] slowly re-fill the reputation of down-weighted backends as they successfully return content (addressed in https://github.com/filecoin-saturn/caboose/pull/26/files by Aarsh).
* [x] handle 'non-existant' content properly - don't downrank nodes when the error is with bad content rather than bad nodes


pool.go
```diff
@@ -161,26 +180,69 @@ func (p *pool) fetchWith(ctx context.Context, c cid.Cid, with string) (blk block
...
-		p.lk.RLock()
-		member := p.c.LocateKey([]byte(aff))
+	p.lk.RLock()
+	if left > len(p.endpoints) {
+		left = len(p.endpoints)
+	}
+	if p.c == nil || p.c.Size() == 0 {
 		p.lk.RUnlock()
-		root := member.String()
+		return nil, ErrNoBackend
```
lidel, line 195: :+1: (we were missing this termination before)

pool.go
```diff
@@ -161,26 +180,69 @@ func (p *pool) fetchWith(ctx context.Context, c cid.Cid, with string) (blk block
...
+		case p.refresh <- struct{}{}:
+		default:
+		}
+
+		return nil, ErrNoBackend
+	}
 
-		blk, err = p.doFetch(ctx, root, c)
+	for i := 0; i < len(nodes); i++ {
+		blk, err = p.doFetch(ctx, nodes[i], c)
```
lidel, line 209: If you have no timeout for `doFetch` here, you effectively have no retries.

In my tests, either saturn PoP replies fast with block, or it hangs until [60s timeout in biforst-gateway](https://github.com/ipfs/bifrost-gateway/pull/33) hits, and cancels the entire context. There is never a scenario when `MaxRetries` logic gets executed here.

Some ideas: 
- if you introduce  timeout per `doFetch` attempt, maybe 19s (so 3 retries take  <60s) (we already downvote on each failure)
- try at least 2 PoPs in parallel, and use the fastest response, cancelling remaining ones (+ boost weight of the fast one)

lidel (changes requested):
Over time, this helps a bit with [60s timeout discussed on Slack](https://filecoinproject.slack.com/archives/C04JQ9X7ENB/p1676379525945819?thread_ts=1676319726.536209&cid=C04JQ9X7ENB).

But retry logic is dead code (see comment below)

aarshkshah1992:
I am taking ownership of this.

pool.go
aarshkshah1992: Addressed in https://github.com/filecoin-saturn/caboose/pull/26/commits/c13c0132bf17f1fd79295c5d15723725eca5d941.

willscott:
the open bullet of 'don't downweight for 404's' is captured in #28 as a follow-on item

willscott:
we should try to fix the flaky test here - i believe it has to do with potential non-determinism in how the consistent hash re-shards on node re-weighting.

tchardin:
@willscott is the bug on this branch or on master?

tchardin:
Sorry I mean is the gateway using this branch or master?

willscott:
gateway is using / hanging bug is on this branch

willscott:
noting that we have made the change such that we can say this branch will:

fix #27

willscott:
fix #16 
fix #25

lidel (approved):
This shipped to `bifrost-stage1-ny`, lgtm.

ps. @willscott automatic closing of issues works only if you add "fix #16 " to the top comment, by editing it

PR #26, opened 2023-02-15
 Fix bugs, refactor, document and land the weighted consistent hashing branch

### LINKED ISSUE #25: Put an upper bound on the number of bytes read when Caboose gets a block from Saturn
Caboose loads the entire block in memory from a potentially malicious L1, either we do a streaming read or reject the response if the block exceeds a certain size.

### LINKED ISSUE #27: Use https:// instead of http://
https://github.com/filecoin-saturn/caboose/blob/93567ae43731920137b20f05a3877a10aeba6008/pool.go#L190 

> You're gonna get 301'd to https on every req

aarshkshah1992:
This PR aims to land https://github.com/filecoin-saturn/caboose/pull/19 and make some changes on top of it. 

The main changes include:

- Take care of `slowly re-fill the reputation of down-weighted backends as they successfully return content` TODO in https://github.com/filecoin-saturn/caboose/pull/19. The way we do this is by bumping up the weights of downvoted nodes that successfully return content by 20 percent on every success with a debounce until they hit the default weight of 20 that we assign to all nodes.

- Addresses @lidel 's comment at https://github.com/filecoin-saturn/caboose/pull/19#discussion_r1105843807 by ensuring we timeout requests to the Saturn L1s with a reasonable default.

- I want to make some changes around how we refresh the pool after getting back a list of Saturn nodes from the Orchestrator and how we read the blocks in memory. Please take a look at https://github.com/filecoin-saturn/caboose/pull/26#discussion_r1107077024 and https://github.com/filecoin-saturn/caboose/pull/26#discussion_r1107080733 and let me know what you think.

- I think I found and fixed a bug around how we were updating the weights. Please see  https://github.com/filecoin-saturn/caboose/pull/26#discussion_r1107075422

- Some refactoring to remove boilerplate and added documentation for people to grok what's happening in the code.

- I'll add some comprehensive unit tests in follow up PRs as there's a lot of code here that needs testing.

Closes #27 Closes #25

### DIFF
--- a/caboose.go
+++ b/caboose.go
@@ -37,9 +37,9 @@ type Config struct {
 	// PoolRefresh is the interval at which we refresh the pool of Saturn nodes.
 	PoolRefresh time.Duration
 
-	// PoolFailureDownvoteDebounce is the amount of time we wait between consecutive updates to the weight of a Saturn node
+	// PoolWeightChangeDebounce is the amount of time we wait between consecutive updates to the weight of a Saturn node
 	// in our pool after a retrieval success/failure.
-	PoolFailureDownvoteDebounce time.Duration
+	PoolWeightChangeDebounce time.Duration
 
 	// trigger early refreshes when pool size drops below this low watermark
 	PoolLowWatermark int
@@ -73,8 +73,8 @@ func NewCaboose(config *Config) (ipfsblockstore.Blockstore, error) {
 			Timeout: DefaultSaturnRequestTimeout,
 		}
 	}
-	if c.config.PoolFailureDownvoteDebounce == 0 {
-		c.config.PoolFailureDownvoteDebounce = DefaultPoolFailureDownvoteDebounce
+	if c.config.PoolWeightChangeDebounce == 0 {
+		c.config.PoolWeightChangeDebounce = DefaultPoolFailureDownvoteDebounce
 	}
 	if c.config.PoolLowWatermark == 0 {
 		c.config.PoolLowWatermark = DefaultPoolLowWatermark
--- a/caboose.go
+++ b/caboose.go
@@ -18,27 +18,27 @@ type Config struct {
 	// OrchestratorClient is the HTTP client to use when communicating with the Saturn orchestrator.
 	OrchestratorClient *http.Client
 
-	// LoggingEndpoint is the URL of the Caboose logging endpoint where we submit the logs pertaining to our Saturn retrieval requests.
+	// LoggingEndpoint is the URL of the logging endpoint where we submit logs pertaining to our Saturn retrieval requests.
 	LoggingEndpoint url.URL
-	// LoggingClient is the HTTP client to use when communicating with the Caboose logging endpoint.
+	// LoggingClient is the HTTP client to use when communicating with the logging endpoint.
 	LoggingClient *http.Client
-	// LoggingInterval is the interval at which we submit logs to the Caboose logging endpoint.
+	// LoggingInterval is the interval at which we submit logs to the logging endpoint.
 	LoggingInterval time.Duration
 
 	// SaturnClient is the HTTP client to use when retrieving content from the Saturn network.
 	SaturnClient *http.Client
 	ExtraHeaders *http.Header
 
-	// DoValidation is used to determine if Caboose should validate the blocks recieved from the Saturn network.
+	// DoValidation is used to determine if we should validate the blocks recieved from the Saturn network.
 	DoValidation bool
-	// If set, AffinityKey is used instead of the block cid as the key on the consistent hashing ring
-	// to decide which Saturn node to retrieve the block from.
+	// If set, AffinityKey is used instead of the block cid as the key on the Saturn node pool
+	// to determine which Saturn node to retrieve the block from.
 	AffinityKey string
 	// PoolRefresh is the interval at which we refresh the pool of Saturn nodes.
 	PoolRefresh time.Duration
 
-	// PoolFailureDownvoteDebounce is the amount of time we wait between consecutive downvotings of a Saturn node
-	// on our consistent hashing ring after a retrieval failure.
+	// PoolFailureDownvoteDebounce is the amount of time we wait between consecutive updates to the weight of a Saturn node
+	// in our pool after a retrieval success/failure.
 	PoolFailureDownvoteDebounce time.Duration
 
 	// trigger early refreshes when pool size drops below this low watermark
--- a/caboose.go
+++ b/caboose.go
@@ -55,8 +55,8 @@ const maxBlockSize = 4194305 // 4 Mib + 1 byte
 const DefaultOrchestratorEndpoint = "https://orchestrator.strn.pl/nodes/nearby?count=1000"
 
 var ErrNotImplemented error = errors.New("not implemented")
-var ErrNoBackend error = errors.New("no available backend")
-var ErrBackendFailed error = errors.New("backend failed")
+var ErrNoBackend error = errors.New("no available strn backend")
+var ErrBackendFailed error = errors.New("strn backend failed")
 
 type Caboose struct {
 	config *Config
--- a/caboose.go
+++ b/caboose.go
@@ -50,6 +50,7 @@ type Config struct {
 const DefaultMaxRetries = 3
 const DefaultPoolFailureDownvoteDebounce = time.Second
 const DefaultPoolLowWatermark = 5
+const DefaultSaturnRequestTimeout = 19 * time.Second
 
 var ErrNotImplemented error = errors.New("not implemented")
 var ErrNoBackend error = errors.New("no available backend")
@@ -68,7 +69,9 @@ func NewCaboose(config *Config) (ipfsblockstore.Blockstore, error) {
 	}
 	c.pool.logger = c.logger
 	if c.config.SaturnClient == nil {
-		c.config.SaturnClient = http.DefaultClient
+		c.config.SaturnClient = &http.Client{
+			Timeout: DefaultSaturnRequestTimeout,
+		}
 	}
 	if c.config.PoolFailureDownvoteDebounce == 0 {
 		c.config.PoolFailureDownvoteDebounce = DefaultPoolFailureDownvoteDebounce
--- a/caboose.go
+++ b/caboose.go
@@ -13,25 +13,38 @@ import (
 )
 
 type Config struct {
+	// OrchestratorEndpoint is the URL of the Saturn orchestrator.
 	OrchestratorEndpoint url.URL
-	OrchestratorClient   *http.Client
+	// OrchestratorClient is the HTTP client to use when communicating with the Saturn orchestrator.
+	OrchestratorClient *http.Client
 
+	// LoggingEndpoint is the URL of the Caboose logging endpoint where we submit the logs pertaining to our Saturn retrieval requests.
 	LoggingEndpoint url.URL
-	LoggingClient   *http.Client
+	// LoggingClient is the HTTP client to use when communicating with the Caboose logging endpoint.
+	LoggingClient *http.Client
+	// LoggingInterval is the interval at which we submit logs to the Caboose logging endpoint.
 	LoggingInterval time.Duration
 
-	Client       *http.Client
+	// SaturnClient is the HTTP client to use when retrieving content from the Saturn network.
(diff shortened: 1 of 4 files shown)


caboose.go
```diff
@@ -13,30 +13,44 @@ import (
...
+	PoolRefresh time.Duration
+
+	// PoolFailureDownvoteDebounce is the amount of time we wait between consecutive updates to the weight of a Saturn node
+	// in our pool after a retrieval success/failure.
 	PoolFailureDownvoteDebounce time.Duration
-	PoolMaxSize                 int
+
 	// trigger early refreshes when pool size drops below this low watermark
 	PoolLowWatermark int
-	MaxConcurrency   int
```
aarshkshah1992, line 33: This is never used.

caboose.go
```diff
@@ -13,30 +13,44 @@ import (
...
+	// If set, AffinityKey is used instead of the block cid as the key on the Saturn node pool
+	// to determine which Saturn node to retrieve the block from.
+	AffinityKey string
+	// PoolRefresh is the interval at which we refresh the pool of Saturn nodes.
+	PoolRefresh time.Duration
+
+	// PoolFailureDownvoteDebounce is the amount of time we wait between consecutive updates to the weight of a Saturn node
+	// in our pool after a retrieval success/failure.
 	PoolFailureDownvoteDebounce time.Duration
-	PoolMaxSize                 int
```
aarshkshah1992, line 30: This is never used.

cmd/caboose/main.go
```diff
@@ -50,15 +50,17 @@ func main1() int {
 
 			cb, err := caboose.NewCaboose(&caboose.Config{
 				OrchestratorEndpoint: *oe,
-				OrchestratorClient:   http.DefaultClient,
+				OrchestratorClient: &http.Client{
```
aarshkshah1992, line 53: @lidel @willscott I don't see us creating a default orchestrator HTTP client in the Caboose constructor. Does the IPFS GW always supply an initialised Orchestrator HTTP client ?

pool.go
```diff
@@ -38,57 +41,66 @@ func (p *pool) loadPool() ([]string, error) {
...
+	if m.lk.TryLock() {
 		if time.Since(m.lastUpdate) > debounce {
 			// make the down-voted member
-			nm := NewMember(m.string)
-			nm.replication = m.replication / 2
-			m.lastUpdate = time.Now()
-			m.Unlock()
+			nm := NewMember(m.url)
+			nm.replication = updateWeightF(m.replication)
+			nm.lastUpdate = time.Now()
```
aarshkshah1992, line 99: We need to update the `lastUpdate` time on this newly constructed member and not on the old member as the pool will now be updated with this member and we want to retain the last updated time across weight updates.

aarshkshah1992: @willscott

pool.go
```diff
@@ -110,12 +122,19 @@ func (p *pool) doRefresh() {
 	newEP, err := p.loadPool()
 	if err == nil {
 		p.lk.RLock()
+
+		// TODO: The orchestrator periodically prunes "bad" L1s based on a reputation system
+		// it owns and runs. We should probably just forget about the Saturn endpoints that were
+		// previously in the pool but are no longer being returned by the orchestrator. It's highly
+		// likely that the Orchestrator has deemed them to be non-functional/malicious.
+		// Let's just override the old pool with the new endpoints returned here.
```
aarshkshah1992, line 130: @willscott @lidel 

The orchestrator periodically tests the L1s to mantain a list of responsive L1s.

pool.go
```diff
@@ -298,15 +368,18 @@ func (p *pool) doFetch(ctx context.Context, from string, c cid.Cid) (b blocks.Bl
...
+	defer resp.Body.Close()
+
 	fb = time.Now()
 	code = resp.StatusCode
 	proto = resp.Proto
 	respReq = resp.Request
-	defer resp.Body.Close()
+
+	// TODO: What if the Saturn node is malicious ? We should have an upper bound on how many bytes we read here.
 	rb, err := io.ReadAll(resp.Body)
```
aarshkshah1992, line 383: @willscott @lidel Is this a concern ?

lidel: If L1 returned HTTP Status Code 200 but the hash does not match, it should be penalized  way more than if it returned with error (code other than 200) or timeouted.

pool.go
lidel: I've asked for access to https://github.com/filecoin-saturn/orchestrator to see what exactly is "checked".

I don't think we can trust Orchestrator to give us usefull L1s, we need to verify ourselves.

The tests they perform are not end-to-end, use hardcoded CIDs, and if the first try fails, they allow cached result (so nodes that lost retrieval capability, but have test CIDs in cache are never penalized):
- https://github.com/filecoin-saturn/orchestrator/blob/04b1f712ef612b340142f3e49e89a995e7e3154e/cron/health-check.js#L27
- https://github.com/filecoin-saturn/orchestrator/blob/04b1f712ef612b340142f3e49e89a995e7e3154e/cron/random-health-check.js#L20-L32

Orchestrator opinion would be useful only if it would run a real IPFS node, add random data to IPFS, announce it on DHT, and then attempt to retrieve it via specific L1. 

This means "nearby" has little value. 
We need to fetch as many L1s we can, and then let the weighting logic to find the most useful ones.
Currently, we only get 25, but with https://github.com/filecoin-saturn/orchestrator/issues/83 we will be able to ask for more.

cmd/caboose/main.go
lidel: We pass our own http.Client for thing like a  proper `User-Agent` header:
https://github.com/ipfs/bifrost-gateway/blob/e0b155e26a963acdcbf2665acd218ace929be0fb/blockstore.go#L80-L81

pool.go
```diff
@@ -207,51 +228,98 @@ func (p *pool) fetchWith(ctx context.Context, c cid.Cid, with string) (blk block
...
-			var nm *Member
-			var needUpdate bool
-			for j, m := range p.endpoints {
-				if m.String() == nodes[i] {
-					if nm, needUpdate = m.Downvote(p.config.PoolFailureDownvoteDebounce); needUpdate {
-						idx = j
+		var idx int
+		var nm *Member
+
+		if err == nil {
```
lidel, line 234: We don't want to lose visibility introduced by https://github.com/filecoin-saturn/caboose/pull/24
```suggestion
		if err != nil {
				goLogger.Debugw("fetch failed", "from", nodes[i], "of", c, "attempt", i, "error", err)
		} else {
```

pool.go
aarshkshah1992: @lidel This make sense to me. I think you'll have to make this change in Bifrost GW as caboose uses whatever orchestrator API URL you pass to it. Caboose Config param you'll have to change is called `OrchestratorEndpoint`.

pool.go
aarshkshah1992: Done.

pool.go
aarshkshah1992: PR update with both these changes.

cmd/caboose/main.go
```diff
@@ -38,7 +38,7 @@ func main1() int {
 			}
 			out := args.Get(1)
 
-			oe, _ := url.Parse("https://orchestrator.strn.pl/nodes/nearby")
+			oe, _ := url.Parse("https://orchestrator.strn.pl/nodes/nearby?count=100")
```
lidel, line 41: This parameter should be moved inside of caboose asa a `const`, appended to the endpoint.
You don't want to ask bifrost-gateway to change URL every time you want to adjust strategy for refreshing L1s :)

pool.go
```diff
@@ -298,31 +372,42 @@ func (p *pool) doFetch(ctx context.Context, from string, c cid.Cid) (b blocks.Bl
...
+	resp, err := p.config.SaturnClient.Do(req)
 	if err != nil {
 		return nil, err
 	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return nil, ErrBackendFailed
+	}
+
 	fb = time.Now()
```
lidel, line 384: We don't want to fake metrics in `goLogger.Debugw("fetch result",`  – surely producing  5XX will take more than 0ms, especially if they timeouted DHT lookup, no?  :upside_down_face: 

```suggestion
	fb = time.Now()
	if resp.StatusCode != http.StatusOK {
		return nil, ErrBackendFailed
	}


```

pool.go
```diff
@@ -298,31 +372,42 @@ func (p *pool) doFetch(ctx context.Context, from string, c cid.Cid) (b blocks.Bl
...
+	received, err = io.ReadFull(resp.Body, buf)
+	// ReadFull returns an EOF ONLY if no bytes were read.
+	if err == io.EOF {
+		return nil, ErrBackendFailed
+	}
+	// If we read less than max bytes, we are good here.
+	if received > 0 && received <= maxBlockSize {
+		buf = buf[:received]
+	} else {
+		return nil, ErrBackendFailed
```
lidel, line 399: This should be a separate error – otherwise we will bang our heads why things do not work N years from now, if ever the  block size policy changes (but fine to do it in https://github.com/filecoin-saturn/caboose/issues/28)

lidel (changes requested):

pool.go
aarshkshah1992: Yeah will do it in #28

aarshkshah1992:
@willscott Test is failing because we moved the Saturn client to use https and looks like it does NOT trust the cert given by the httptest TLS server.

lidel:
@aarshkshah1992  I think this is why we have https://github.com/ipfs/bifrost-gateway/blob/main/blockstore.go#L82-L84

If you move this hack into caboose it should fix your cmd and help us clean up Saturn-related hack from bifrost-gateway 
(but still apply the outer RoundTrripper, so user-agent override done in bifrost-gatewat still works)

pool.go
willscott: i think change it on both - if a couple things time out simultaneously against the previous member, we want only one of them to do the update re-weighting

pool.go
willscott: i think mostly this was: what happens if we get an empty list from the orchestrator because it has a bug / looses it's DB?
it doesn't seem to hurt much to keep using known, working nodes.
maybe we ask for some way for the orchestrator to more forcefully tell us to replace?

willscott (approved):

lidel:
@aarshkshah1992  I've fixed `go test` in https://github.com/filecoin-saturn/caboose/pull/26/commits/d2749d3b8402a574db51b42d311eccc75e12738f, will now take it for a spin in bifrost-gateway.

pool.go
lidel: We could be asking for `?count=9999` and see all of them, and decrease weight of ones that were removed.
This way if  they return empty list, of only 4 for some reason, we are not left with broken client – weights will be applied to all, so no impact.

lidel (approved):
[233eb8a](https://github.com/ipfs/bifrost-gateway/commit/233eb8a9df0a8bd3fcfb75b9f972b499eac278d5) was able to retrieve `bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi`, I am switching bifrost-gateway to this commit: https://github.com/ipfs/bifrost-gateway/pull/39

pool.go
```diff
@@ -208,50 +229,100 @@ func (p *pool) fetchWith(ctx context.Context, c cid.Cid, with string) (blk block
...
+		p.lk.RUnlock()
+
+		// we weren't able to downvote the failing Saturn node, let's just retry the fetch with a new node.
+		if idx == -1 || nm == nil {
+			continue
+		}
+
+		// we need to take the lock as we're updating the list of pool endpoint members below.
+		p.lk.Lock()
+		if p.endpoints[idx].url == nm.url {
```
lidel, line 279: :warning:  @aarshkshah1992 mind checking if my understanding is correct that this line is causing random panics? 
Details in  https://github.com/filecoin-saturn/caboose/issues/29

willscott: i think this is because between the `Runlock` and the `Lock` the `endpoints` can change to a new set. we should find `idx` within the same critical section for stability

willscott: #30 should fix

PR #32, opened 2023-02-19
add latency / size metrics for error cases

willscott:
report these even in error cases

### DIFF
--- a/pool.go
+++ b/pool.go
@@ -349,7 +349,7 @@ func (p *pool) doFetch(ctx context.Context, from string, c cid.Cid) (b blocks.Bl
 	requestId := uuid.NewString()
 	goLogger.Debugw("doing fetch", "from", from, "of", c, "requestId", requestId)
 	start := time.Now()
-	fb := time.Now()
+	fb := time.Unix(0, 0)
 	code := 0
 	proto := "unknown"
 	respReq := &http.Request{}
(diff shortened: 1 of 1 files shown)


pool.go
```diff
@@ -345,21 +342,25 @@ func (p *pool) updateWeightUnlocked(node string, failure bool) (index int, membe
...
+	fb := time.Unix(0, 0)
 	code := 0
 	proto := "unknown"
 	respReq := &http.Request{}
 	received := 0
 	defer func() {
-		goLogger.Debugw("fetch result", "from", from, "of", c, "status", code, "size", received, "ttfb", int(fb.Sub(start).Milliseconds()), "duration", time.Since(start).Seconds())
+		ttfbMs := fb.Sub(start).Milliseconds()
+		durationSecs := time.Since(start).Seconds()
+		goLogger.Debugw("fetch result", "from", from, "of", c, "status", code, "size", received, "ttfb", int(ttfbMs), "duration", durationSecs, "attempt", attempt, "error", e)
```
lidel, line 357: :information_source:  replaced  two debugs logs per request with one (success will have error equal null here)

lidel (approved):
lgtm (it was bit hairy so cleaned it up, and deduplicated calculation between logs and metrics)


ps. I think "Speed" one may not be correctly represented at [bifrost-gw-staging-metrics](https://protocollabs.grafana.net/d/d6FxxgJ4k/bifrost-gw-staging-metrics?orgId=1&from=now-24h&to=now&refresh=1m), Y-axis should be something like bytes/second, and not seconds(?). (will update staging to  https://github.com/ipfs/bifrost-gateway/commit/e4d22d2b9aa7eecae607ffa9ab6d2837262ff5f0 binary with this PR, and we will ping George later this week if its still wonky)

Issue #28, opened 2023-02-16
Downvoting of Saturn L1s should consider the return codes from L1s


aarshkshah1992:
If L1 is simply unavailable/returning bad data, then it should be a hard downvote,
If L1 is unable to find some content, then it should be a soft downvote.

Depends on https://github.com/filecoin-saturn/L1-node/issues/222 as L1s need to start returning granular status codes for this to happen.

PR #33, opened 2023-02-20
Refactor locks to event loop and feed Saturn HTTP status codes into the weights

aarshkshah1992:
The ring weights locking model was getting hard to reason about.
This PR refactors it to use an event loop instead.
Also, closes #https://github.com/filecoin-saturn/caboose/issues/28.

### DIFF
--- a/caboose.go
+++ b/caboose.go
@@ -58,6 +58,9 @@ var ErrNotImplemented error = errors.New("not implemented")
 var ErrNoBackend error = errors.New("no available strn backend")
 var ErrBackendFailed error = errors.New("strn backend failed")
 
+// ErrTransient is returned when a transient error(timeouts, not found etc) occurs while fetching from the strn backend.
+var ErrTransient error = errors.New("transient error while fetching from strn backend")
+
 type Caboose struct {
 	config *Config
 	pool   *pool
--- a/caboose.go
+++ b/caboose.go
@@ -48,7 +48,7 @@ type Config struct {
 }
 
 const DefaultMaxRetries = 3
-const DefaultPoolFailureDownvoteDebounce = time.Second
+const DefaultPoolFailureDownvoteDebounce = 2 * time.Second
 const DefaultPoolLowWatermark = 5
 const DefaultSaturnRequestTimeout = 19 * time.Second
 const maxBlockSize = 4194305 // 4 Mib + 1 byte
--- a/failure_test.go
+++ b/failure_test.go
@@ -58,7 +58,7 @@ func TestCabooseFailures(t *testing.T) {
 		DoValidation:             false,
 		PoolWeightChangeDebounce: time.Duration(1),
 		PoolRefresh:              time.Millisecond * 50,
-		MaxRetrievalAttempts:     2,
+		MaxRetrievalAttempts:     3,
 	})
 	if err != nil {
 		t.Fatal(err)
--- a/pool.go
+++ b/pool.go
@@ -39,24 +39,51 @@ func (p *pool) loadPool() ([]string, error) {
 	return responses, nil
 }
 
+type updateRingMembersReq struct {
+	newMembers []string
+	done       chan struct{}
+}
+
+type getRingNodesForKeyReq struct {
+	key  string
+	n    int
+	resp chan getRingNodesResp
+}
+
+type getRingNodesResp struct {
+	nodes []string
+	ok    bool
+	err   error
+}
+
 type pool struct {
+	closeOnce sync.Once
+	ctx       context.Context
+	cancel    context.CancelFunc
+	wg        sync.WaitGroup
+
 	config *Config
 	logger *logger
 
 	started chan struct{} // started signals that we've already initialized the pool once with Saturn endpoints.
 	refresh chan struct{} // refresh is used to signal the need for doing a refresh of the Saturn endpoints pool.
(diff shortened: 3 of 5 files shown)


failure_test.go
```diff
@@ -58,7 +58,7 @@ func TestCabooseFailures(t *testing.T) {
 		DoValidation:             false,
 		PoolWeightChangeDebounce: time.Duration(1),
 		PoolRefresh:              time.Millisecond * 50,
-		MaxRetrievalAttempts:     2,
+		MaxRetrievalAttempts:     3,
```
aarshkshah1992, line 61: This is necessary so that the one remaining good backend always gets used in line 144 at https://github.com/filecoin-saturn/caboose/blob/ebf9e25e4a94102072b215c4e7632da20fe74c15/failure_test.go#L142

Otherwise, it sometimes picks the two "invalid backends" for the key and test becomes flaky.

caboose.go
```diff
@@ -58,6 +58,9 @@ var ErrNotImplemented error = errors.New("not implemented")
 var ErrNoBackend error = errors.New("no available strn backend")
 var ErrBackendFailed error = errors.New("strn backend failed")
 
+// ErrTransient is returned when a transient error(timeouts, not found etc) occurs while fetching from the strn backend.
+var ErrTransient error = errors.New("transient error while fetching from strn backend")
```
willscott, line 62: what is the consumer (e.g. `bifrost-gateway`) supposed to do when it gets this error?
* if this is returned to consumer, i'd expect it to wrap the underlying error rather than replacing it
* if it's meant to be purely internal, it shouldn't be exported

pool.go
```diff
@@ -39,24 +43,62 @@ func (p *pool) loadPool() ([]string, error) {
...
+type updateRingMembersReq struct {
+	newMembers []string
+	done       chan struct{}
+}
+
+type getRingNodesReq struct {
+	key  string
+	n    int
+	resp chan getRingNodesResp
+}
```
willscott, line 55: I'm worried about the scaling limitation that would happen if standard gets are synchronized this way.

in the current design, the standard case is that each inbound request takes an `RLock`, which many threads can do without a point of synchronization/serialization. By making every in-bound request ask for nodes from a single event-loop thread you're serializing all requests against what a single core / event-loop go-routine has time for.

pool.go
```diff
@@ -191,152 +196,270 @@ func (p *pool) refreshPool() {
...
-	aff := with
-	if aff == "" {
-		aff = c.Hash().B58String()
+func (p *pool) ringLoop() {
+	defer p.wg.Done()
+
+	successKeysCache := cache.New(5*time.Minute, 2*time.Minute)
+	transientErrorsCache := cache.New(5*time.Minute, 2*time.Minute)
+
+	downVoteFnc := func(node string, pct int) {
```
willscott, line 218: why is this defined as an anonymous function and not as a method on pool?

pool.go
```diff
@@ -39,24 +43,62 @@ func (p *pool) loadPool() ([]string, error) {
...
 	started chan struct{} // started signals that we've already initialized the pool once with Saturn endpoints.
 	refresh chan struct{} // refresh is used to signal the need for doing a refresh of the Saturn endpoints pool.
-	done    chan struct{} // done is used to signal that we're shutting down the Saturn endpoints pool and don't need to refresh it anymore.
 
-	lk        sync.RWMutex
-	endpoints MemberList         // guarded by lk
-	c         *hashring.HashRing // guarded by lk
+	// ringLoop vars
+	endpoints MemberList         // guarded by ringLoop
+	c         *hashring.HashRing // guarded by ringLoop
```
willscott, line 88: why aren't these just local variables in ringLoop?

pool.go
```diff
@@ -77,99 +117,64 @@ type Member struct {
...
+		getRingNodesCh:      make(chan getRingNodesReq),
+		upvoteNodeCh:        make(chan upvoteReq, 256),
+		downvoteNodeCh:      make(chan downVoteReq, 256),
 	}
+
+	p.wg.Add(1)
 	go p.refreshPool()
+
+	p.wg.Add(1)
+	go p.ringLoop()
```
willscott, line 149: what are the semantics of `wg`?

pool.go
```diff
@@ -191,152 +196,270 @@ func (p *pool) refreshPool() {
...
+	successKeysCache := cache.New(5*time.Minute, 2*time.Minute)
+	transientErrorsCache := cache.New(5*time.Minute, 2*time.Minute)
+
+	downVoteFnc := func(node string, pct int) {
+		for i, m := range p.endpoints {
+			if m.url == node {
+				if time.Since(m.lastUpdate) > p.config.PoolWeightChangeDebounce {
+					m.lastUpdate = time.Now()
+
+					m.replication = m.replication * ((100 - pct) / 100)
```
willscott, line 224: integer rounding is going to make this not always do what we want, right?

willscott (commented):
In general i think my main hesitation is that i find the structure, and in particular the complexity of `ringLoop` you're introducing (a 175 line method) to be as hard to work with as the current lock structure.

i wonder if factoring out a couple functions from `fetchWith` that represent the scopes where a lock is held would make things clear enough. the overloading of that method to do both retries and weight updates seems like much of what makes things tricky at the moment.

PR #36, opened 2023-02-20
Factor out repeated code to simplify locking, reduce lock granularity and add membership debounce and tests

willscott:
This is a counter to #33 

Rather than adding an event loop, adding a method for `replaceNodeToHaveWeight` leads to an overall code reduction and less potential for missed locking semantics.

@aarshkshah1992  added debounce for membership and tests.


pool.go
```diff
@@ -240,87 +240,83 @@ func (p *pool) fetchWith(ctx context.Context, c cid.Cid, with string) (blk block
...
+	var idx int
+	var nm *Member
+
+	if err == nil {
+		// Saturn fetch worked, we should try upvoting.
+		p.lk.RLock()
+		idx, nm = p.updateWeightUnlocked(node, false)
+		p.lk.RUnlock()
+
+		if idx != -1 && nm != nil && p.endpoints[idx].url == nm.url {
```
aarshkshah1992, line 300: Shouldn't we check the `p.endpoints[idx].url == nm.url` thing under a read lock ?

aarshkshah1992: Fixed this in https://github.com/filecoin-saturn/caboose/pull/39

willscott: good catch 🙏

willscott:
I think this looks good.

@lidel : want to try a build against this branch before we merge to main?

lidel (approved):
lgtm

i did basic `ab -k -n 20000 -c 1000 -w "http://en.wikipedia-on-ipfs.org.ipns.localhost:8081/wiki/Archaeology"` with `BLOCK_CACHE_SIZE=2 bifrost-gateway` and nothing crashed nor locked.

It has surprisingly low number of fetch errors too, which is :+1: 

ps. realized we had unnecessary buckets for durations below 50ms (even on localhost test), pushed fix to  account for that.

PR #37, opened 2023-02-21
add a metric of pool health

### LINKED ISSUE #35: Expose peer pool health as metric
Caboose is running weighting algorith introduced in #19 and already tracks the `pool_size`

Thoughts on adding second metric (histogram?) that tracks number of peers with positive vs negative weights?

This would allow us to track pool health over time, identify causes of things like perf degradation, or spikes in certain hours.

cc @willscott  @gmasgras for visibility

willscott:
fix #35

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -17,6 +17,11 @@ var (
 		Help: "Number of active caboose peers",
 	})
 
+	poolHealthMetric = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: prometheus.BuildFQName("ipfs", "caboose", "pool_health"),
+		Help: "Helath of the caboose pool",
+	}, []string{"weight"})
+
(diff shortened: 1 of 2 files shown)


metrics.go
```diff
@@ -17,6 +17,11 @@ var (
 		Help: "Number of active caboose peers",
 	})
 
+	poolHealthMetric = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: prometheus.BuildFQName("ipfs", "caboose", "pool_health"),
+		Help: "Helath of the caboose pool",
```
aarshkshah1992, line 22:
```suggestion
		Help: "Health of the caboose pool",
```

aarshkshah1992 (approved):
lgtm with one change

PR #39, opened 2023-02-21
Consider error codes in downvoting and build test harness for pool

### LINKED ISSUE #28: Downvoting of Saturn L1s should consider the return codes from L1s

aarshkshah1992:
fixes #28 with tests.

Introduces a test harness for the pool.


pool.go
```diff
@@ -297,27 +306,36 @@ func (p *pool) fetchAndUpdate(ctx context.Context, node string, c cid.Cid, attem
 		idx, nm = p.updateWeightUnlocked(node, false)
 		p.lk.RUnlock()
 
-		if idx != -1 && nm != nil && p.endpoints[idx].url == nm.url {
+		if idx != -1 && nm != nil {
```
aarshkshah1992, line 309: endpoints was being read outside a lock.

pool.go
```diff
@@ -126,10 +127,14 @@ func newPool(c *Config) *pool {
...
 	}
-	go p.refreshPool()
+
 	return &p
 }
 
+func (p *pool) Start() {
+	go p.refreshPool()
+}
+
```
willscott, line 137: what do we get by factoring this out as a separate API to call versus having it implicit in `newPool`?

pool.go
```diff
@@ -282,42 +296,50 @@ func (p *pool) replaceNodeToHaveWeight(nm *Member) {
...
 
 	// we weren't able to downvote the node.
 	if idx == -1 || nm == nil {
 		return
 	}
 
-	// we need to take the lock as we're updating the list of pool endpoint members below.
 	p.replaceNodeToHaveWeight(nm)
-	return
 }
```
willscott, line 343: can `upvote` and `downvote` be one method with `direction bool` since it seems like they basically just differ in the `updateWeightUnlocked` bool

pool.go
```diff
@@ -282,42 +296,50 @@ func (p *pool) replaceNodeToHaveWeight(nm *Member) {
...
-		}
-
+		p.upvote(node)
 		// Saturn fetch worked, we return the block.
 		return
 	}
 
+	// If this is a NOT found or Timeout error, park the downvoting for now and see if other members are able to give us this content.
+	if errors.Is(err, ErrContentNotFound) || errors.Is(err, ErrSaturnTimeout) {
+		transientErrs[node] = struct{}{}
```
willscott, line 313: instead of `struct{}{}` maybe worth tracking the `err` itself?

pool.go
```diff
@@ -240,9 +248,14 @@ func (p *pool) fetchWith(ctx context.Context, c cid.Cid, with string) (blk block
...
 
 	for i := 0; i < len(nodes); i++ {
-		blk, err = p.fetchAndUpdate(ctx, nodes[i], c, i)
+		blk, err = p.fetchAndUpdate(ctx, nodes[i], c, i, transientErrs)
 
 		if err == nil {
+			// downvote all parked failed nodes as some other node was able to give us the required content here
+			for node := range transientErrs {
+				p.downvote(node)
+			}
```
willscott, line 257: how much complexity would it introduce to have this batch of operations occur within one lock acquisition? (i'm okay for this to be a subsequent PR)

willscott (commented):
This looks generally good

caboose.go
```diff
@@ -53,10 +53,13 @@ const DefaultPoolLowWatermark = 5
 const DefaultSaturnRequestTimeout = 19 * time.Second
 const maxBlockSize = 4194305 // 4 Mib + 1 byte
 const DefaultOrchestratorEndpoint = "https://orchestrator.strn.pl/nodes/nearby?count=1000"
+const DefaultPoolRefreshInterval = 5 * time.Minute
 
 var ErrNotImplemented error = errors.New("not implemented")
 var ErrNoBackend error = errors.New("no available strn backend")
 var ErrBackendFailed error = errors.New("strn backend failed")
+var ErrContentNotFound error = errors.New("content not found")
```
lidel, line 61: What Lassie is doing (returning 404) is semantically incorrect, will cause us headache when we will do GETs for graphs.
Let's make this clear that it is failure to find providers, and not failure to find path in requested path:

```suggestion
var ErrContentProviderNotFound error = errors.New("strn failed to find content providers")
```

I know this looks like a cosmetic difference, but trust me, one day, this will save someone hours in debugging :)

pool.go
aarshkshah1992: @willscott The flexibility to start the refresh loop whenever we want in tests instead of always having to do it on creation of the pool.

pool.go
aarshkshah1992: Hey, have done this. Please take a look.

caboose.go

pool.go
```diff
@@ -267,6 +287,52 @@ func (p *pool) replaceNodeToHaveWeight(nm *Member) {
...
+	p.lk.Lock()
+	defer p.lk.Unlock()
+
+	for _, req := range reqs {
+		idx, nm := p.updateWeightUnlocked(req.node, req.failure)
+		// we weren't able to change the weight.
+		if idx == -1 || nm == nil {
+			continue
+		}
+		p.updatePoolWithNewWeightUnlocked(nm, idx)
```
willscott, line 331: the `UpdateWithWeights` call ends up being expensive - it re-builds the full consistent hash table.

we can do it in a followup PR, but we should see if we can end up making 1 call to `UpdateWithWeights' after applying all our weight changes to `endpoints`.

willscott (approved):

Issue #38, opened 2023-02-21
Saturn logging fixes


guanzo:
There's a few issues that need to be fixed for Saturn logging.

* The URL should be the URL that was requested. Currently it's an ip address
https://github.com/filecoin-saturn/caboose/blob/ebf9e25e4a94102072b215c4e7632da20fe74c15/pool.go#L368

  It should be this template with values filled in.

  https://github.com/filecoin-saturn/caboose/blob/ebf9e25e4a94102072b215c4e7632da20fe74c15/pool.go#L342
  
* Rename `localTime` to `startTime`
   https://github.com/filecoin-saturn/caboose/blob/ebf9e25e4a94102072b215c4e7632da20fe74c15/pool.go#L369

* Rename `requestDuration` to `requestDurationSec`
   https://github.com/filecoin-saturn/caboose/blob/ebf9e25e4a94102072b215c4e7632da20fe74c15/pool.go#L372

* Log is missing a `nodeId` field. The value is obtained from the `Saturn-Node-Id` response header.

* `requestId` should be obtained from the `Saturn-Transfer-Id` response header instead of generating a UUID.
  https://github.com/filecoin-saturn/caboose/blob/ebf9e25e4a94102072b215c4e7632da20fe74c15/pool.go#L346
  
* `cacheHit` should be obtained from the `Saturn-Cache-Status` response header instead of hardcoding to false. 
  Check if the response header value equals `HIT`.
  https://github.com/filecoin-saturn/caboose/blob/ebf9e25e4a94102072b215c4e7632da20fe74c15/pool.go#L367

* `numBytesSent` should include the size of the headers.
  https://github.com/filecoin-saturn/caboose/blob/ebf9e25e4a94102072b215c4e7632da20fe74c15/pool.go#L370-L371

* `clientAddress` and `userAgent` do not need to be set in the logs as the receiving server can obtain both fields from the request.
  https://github.com/filecoin-saturn/caboose/blob/ebf9e25e4a94102072b215c4e7632da20fe74c15/pool.go#L378

* If a range request was sent, the value should have the same format as the `Range` request header ( `bytes=<start>-<end>`). Otherwise an empty string is fine.
  https://github.com/filecoin-saturn/caboose/blob/ebf9e25e4a94102072b215c4e7632da20fe74c15/pool.go#L379

[...]

PR #40, opened 2023-02-21
Send correct logs to Saturn logger 

aarshkshah1992:
Fixes https://github.com/filecoin-saturn/caboose/issues/38.

- [x] Investigate with @guanzo why Saturn logger is returning 400 with the correct format and JWT.
- [x] @guanzo to share the testnet and mainnet logging URLs and JWT with @lidel.
- [x] Actually see a 200 from the testnet and mainnet logging servers before we merge this branch.

### DIFF
--- a/caboose.go
+++ b/caboose.go
@@ -45,9 +45,6 @@ type Config struct {
 	PoolLowWatermark int
 	// MaxRetrievalAttempts determines the number of times we will attempt to retrieve a block from the Saturn network before failing.
 	MaxRetrievalAttempts int
-
-	// SaturnLoggerJWT is the JWT Auth token to use when submitting logs to the Saturn logging endpoint.
-	SaturnLoggerJWT string
 }
 
 const DefaultMaxRetries = 3
@@ -68,16 +65,13 @@ type Caboose struct {
 }
 
 func NewCaboose(config *Config) (ipfsblockstore.Blockstore, error) {
-	if config.SaturnLoggerJWT == "" || len(config.SaturnLoggerJWT) == 0 {
-		return nil, errors.New("JWT token required for Saturn Logger")
-	}
-
 	c := Caboose{
 		config: config,
 		pool:   newPool(config),
 		logger: newLogger(config),
 	}
 	c.pool.logger = c.logger
+
 	if c.config.SaturnClient == nil {
 		c.config.SaturnClient = &http.Client{
 			Timeout: DefaultSaturnRequestTimeout,
--- a/caboose.go
+++ b/caboose.go
@@ -45,6 +45,9 @@ type Config struct {
 	PoolLowWatermark int
 	// MaxRetrievalAttempts determines the number of times we will attempt to retrieve a block from the Saturn network before failing.
 	MaxRetrievalAttempts int
+
+	// SaturnLoggerJWT is the JWT Auth token to use when submitting logs to the Saturn logging endpoint.
+	SaturnLoggerJWT string
 }
 
 const DefaultMaxRetries = 3
@@ -65,6 +68,10 @@ type Caboose struct {
 }
 
 func NewCaboose(config *Config) (ipfsblockstore.Blockstore, error) {
+	if config.SaturnLoggerJWT == "" || len(config.SaturnLoggerJWT) == 0 {
+		return nil, errors.New("JWT token required for Saturn Logger")
+	}
+
 	c := Caboose{
 		config: config,
 		pool:   newPool(config),
--- a/cmd/caboose/main.go
+++ b/cmd/caboose/main.go
@@ -56,9 +56,10 @@ func main1() int {
 				LoggingClient:   http.DefaultClient,
 				LoggingInterval: 5 * time.Second,
 
-				DoValidation: true,
-				PoolRefresh:  5 * time.Minute,
-				SaturnClient: &saturnClient,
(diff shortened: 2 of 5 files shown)


cmd/caboose/main.go
```diff
@@ -84,5 +85,6 @@ func main1() int {
 		log.Println(err)
 		return 1
 	}
+	time.Sleep(5 * time.Second)
```
willscott, line 88: why do we need this?

log.go
```diff
@@ -59,7 +62,24 @@ func (l *logger) submit(logs []log) {
...
+	req, err := http.NewRequest(http.MethodPost, l.endpoint.String(), finalLogs)
+	if err != nil {
+		goLogger.Errorw("failed to create http request to submit saturn logs", "err", err)
+		return
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", l.jwt))
+
+	resp, err := l.client.Do(req)
```
willscott, line 75: rather than having this a specific `saturn logging needs a jwt bearer token` - which is a pretty different level of abstraction than the rest of this library so far - i wonder if we can have this header set by the configured http client.

so e.g. we set the specific client/url at https://github.com/ipfs/bifrost-gateway/blob/main/blockstore.go#L111
we could do something like the final code block at https://developer20.com/add-header-to-every-request-in-go/ to configure the JWT token in the same place, which seems more local than splitting it between there and here.

caboose.go
```diff
@@ -65,6 +68,10 @@ type Caboose struct {
 }
 
 func NewCaboose(config *Config) (ipfsblockstore.Blockstore, error) {
+	if config.SaturnLoggerJWT == "" || len(config.SaturnLoggerJWT) == 0 {
+		return nil, errors.New("JWT token required for Saturn Logger")
+	}
```
lidel, line 73: :warning:  this breaks flow for local development OR requires everyone on the team to get their own token.
iiurc last time we talked about this @guanzo  noted it is ok to disable sending logs when JWT token is not set.

pool.go
```diff
@@ -363,28 +377,45 @@ func (p *pool) doFetch(ctx context.Context, from string, c cid.Cid, attempt int)
...
 			fetchSizeMetric.Observe(float64(received))
 		}
+
+		if respHeader != nil {
+			saturnNodeId = respHeader.Get(saturnNodeIdKey)
+			saturnTransferId = respHeader.Get(saturnTransferIdKey)
+
+			cacheHit := respHeader.Get(saturnCacheHitKey)
+			if cacheHit == saturnCacheHit {
+				isCacheHit = true
```
lidel, line 387: :thought_balloon:  is the idea here to use this and TTFBMS to identify fraud when L1 lied it did the work, but in fact it did no work and returned the thing it already had cached?

cmd/caboose/main.go
aarshkshah1992: This is to wait for some time to ensure we actually see a 200 from the log server as a log line. I'll remove it.

pool.go
aarshkshah1992: @guanzo Can you take a look at this comment ? Why does the L1 need this field ?

guanzo: The idea here is to report if a request was a cache hit or not. This is useful for metrics and payment calculations. What lidel described is a more advanced fraud detection use case that isn't implemented currently

pool.go
```diff
@@ -401,8 +432,10 @@ func (p *pool) doFetch(ctx context.Context, from string, c cid.Cid, attempt int)
 	resp, err := p.config.SaturnClient.Do(req)
 	fb = time.Now()
```
guanzo, line 433: This records the time when headers arrive, right? Let's instead record the time when the first bytes of the body arrive. That's how the other retrieval clients measure TTFB and it's harder for the L1 to cheat that.

guanzo (changes requested):

cmd/caboose/main.go
aarshkshah1992: Removed.

log.go
aarshkshah1992: I have removed this token from the config.

caboose.go

pool.go
aarshkshah1992: Done.

aarshkshah1992:
Filed https://github.com/ipfs/bifrost-gateway/issues/51 for Bifrost to make the required changes.

aarshkshah1992:
@willscott @lidel Don't worry about the failing CI test here. Has already been fixed in https://github.com/filecoin-saturn/caboose/pull/39.

pool.go
lidel: Let's continue in https://filecoinproject.slack.com/archives/C031K3BG527/p1677088893305269 ;-)

guanzo (approved):

cmd/caboose/main.go
lidel: @aarshkshah1992  but.. did you push the removal? :upside_down_face:  I still see it :D

aarshkshah1992:
@willscott Merging this as there's some config work that needs to happen on Bifrost to test this end -end. Work captured at https://github.com/ipfs/bifrost-gateway/issues/51.

PR #46, opened 2023-02-24
feat: gateway.ContentPathKey as affinity key

hacdias:
<del>Context keys should not be strings: https://pkg.go.dev/context#WithValue. I'm not sure whether changing it to `interface{}` - to allow for custom types - will prevent bugs, but I hope so.</del> Refactored it so we dont use `interface{}` – @lidel


caboose.go
```diff
@@ -131,6 +137,25 @@ func (c *Caboose) GetSize(ctx context.Context, it cid.Cid) (int, error) {
...
+		return contentPath
+
+		/* TODO: if we ever want to revisit, and group per root CID of entire DAG:
+		const contentRootIdx = 2
+		if parts := strings.Split(contentPath, "/"); len(parts) > contentRootIdx {
+			// use top level contentRoot ('id' from /ipfs/id or /ipns/id) as affinity key
+			return parts[contentRootIdx]
+		}
+		*/
+	}
```
lidel, line 158: :information_source:  @willscott @aschmahmann  FYI

lidel (approved):

willscott (approved):

Issue #48, opened 2023-02-24
L1 scoring ends up with no peers on bifrost-stage1-ny


lidel:
## Problem

After deploying bifrost-gateway 2023-02-24-c305b3b (with #36) to bifrost-stage1-ny  traffic died very quick:

> ![2023-02-24-234420_3440x1440_scrot](https://user-images.githubusercontent.com/157609/221311663-b4ffb612-805c-48d5-a8ff-2ead0657bfa8.png)

At the same time, Nginx  that is in front of bifrost-gateway staryed returning 502 for nearly every request:

> ![Screenshot 2023-02-24 at 23-57-49 bifrost-gw staging metrics - Bifrost - Dashboards - Grafana](https://user-images.githubusercontent.com/157609/221316473-5121bda0-632d-42fb-8d54-6858e55e31ed.png)

Checked nginx logs and this happen because caboose has `no available strn backend`:
```
> GET /ipfs/QmSr3uQLnXRTfhrpwcNTs14pfphf6yHDHXq4CD8MRBC7mf/17_036.jpg HTTP/1.1
< HTTP/1.1 502 Bad Gateway
failed to resolve /ipfs/QmSr3uQLnXRTfhrpwcNTs14pfphf6yHDHXq4CD8MRBC7mf/17_036.jpg: Bad Gateway: no available strn backend
```

## Some ideas

- I was not able to reproduce this locally, but i tested with `ab -k -n 10000 -c 1000 -w "http://en.wikipedia-on-ipfs.org.ipns.localhost:8081/wiki/Archaeology"` and `BLOCK_CACHE_SIZE=2 ./bifrost-gateway`, which is easy to find, and does not test diverse set of CIDs that would penalize entire pool
-  we are still running against  https://orchestrator.strn.pl/nodes/nearby?count=1000&core=true which returns only ~40 L1s. It could be that given very diverse traffic, we quickly run  out of useful peers.

@willscott @aarshkshah1992 as you are more familiar with  #36

willscott:
@aarshkshah1992  - I hacked around this to get traffic flowing over the weekend for metrics. the compromise is described in #49 

We should figure out how to get the weighting algorithm tuned in.

* maybe test using the `ab` (apache benchmark) client against actual saturn to understand actual failure rate and tune the weighting changes to work for those rates.
* maybe think about up/down voting in relation to the observed 'average' failure rate caboose is currently seeing (#50)

aarshkshah1992:
[...]

PR #49, opened 2023-02-24
keep nodes even if bad. stronger up-weighting

willscott:
Without digging deep into error logs, this updates two of the most likely places that could lead to #48 

* for now, don't populate the `removedTimeCache` so that nodes given by the orchestrator stay considered even if we'd previously decided they were failing
* if a node is partially failing, don't restrict the 'upvotes' to the same schedule as the 'downvotes', so a node that mostly is succeeding will stay in consideration.


lidel (approved):
Deployed `bifrost-gateway 2023-02-24-95c903b` with this hotfix to  `bifrost-stage1-ny`
Data flow is restored :+1:

aarshkshah1992:
Should be taken care of by https://github.com/filecoin-saturn/caboose/pull/59.

We should close this PR once #59 lands.

aarshkshah1992:
Superseded by #59

PR #53, opened 2023-02-27
Use highly ranked backup nodes to restore and recover pool

aarshkshah1992:
Counter to https://github.com/filecoin-saturn/caboose/pull/49.

We should keep the membership cool down cache around to prevent underperforming nodes from getting into the pool. In the event that we aren't able to fill up the pool with enough members, we will use a cache of recent well performing nodes as backup nodes to restore the pool.

### DIFF
--- a/caboose.go
+++ b/caboose.go
@@ -65,7 +65,7 @@ type Config struct {
 	// before we start making retrieval attempts for it.
 	CidCoolDownDuration time.Duration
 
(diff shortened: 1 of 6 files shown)


willscott:
I wonder if this is enough -
we were seeing on friday that all of the nodes were getting down-weighted out of the pool. that means that if we move to this alone we'll end up just always running on the 20 backup nodes.

Do we get anything useful from running in that state this week?

aarshkshah1992:
Closing till I come up with and test the dynamic weighing PR.

PR #60, opened 2023-03-06
feat: common .RetryAfter() interface for errors

hacdias:
Removes the `RetryAfter` property from `ErrSaturnTooManyRequests` and `ErrCidCoolDown` and wraps them with `gateway.ErrRetryAfter`.

### DIFF
--- a/caboose.go
+++ b/caboose.go
@@ -104,25 +104,31 @@ var ErrSaturnTimeout error = errors.New("saturn backend timed out")
 
 type ErrSaturnTooManyRequests struct {
 	Node       string
-	RetryAfter time.Duration // TODO: DRY refactor after https://github.com/ipfs/go-libipfs/issues/188
+	retryAfter time.Duration
 }
 
-func (e ErrSaturnTooManyRequests) Error() string {
-	return fmt.Sprintf("saturn node %s returned Too Many Requests error, please retry after %s", e.Node, humanRetry(e.RetryAfter))
+func (e *ErrSaturnTooManyRequests) Error() string {
+	return fmt.Sprintf("saturn node %s returned Too Many Requests error, please retry after %s", e.Node, humanRetry(e.retryAfter))
+}
+
+func (e *ErrSaturnTooManyRequests) RetryAfter() time.Duration {
+	return e.retryAfter
 }
 
 type ErrCoolDown struct {
 	Cid        cid.Cid
 	Path       string
-	RetryAfter time.Duration // TODO: DRY refactor after https://github.com/ipfs/go-libipfs/issues/188
+	retryAfter time.Duration
 }
 
 func (e *ErrCoolDown) Error() string {
-	return fmt.Sprintf("multiple saturn retrieval failures seen for CID %s/Path %s, please retry after %s", e.Cid, e.Path, humanRetry(e.RetryAfter))
+	return fmt.Sprintf("multiple saturn retrieval failures seen for CID %s/Path %s, please retry after %s", e.Cid, e.Path, humanRetry(e.retryAfter))
 }
 
-// TODO: move this to upstream error interface in https://github.com/ipfs/go-libipfs/issues/188
-// and refactor ErrCoolDown and ErrSaturnTooManyRequests to inherit from that instead
+func (e *ErrCoolDown) RetryAfter() time.Duration {
+	return e.retryAfter
+}
+
 func humanRetry(d time.Duration) string {
 	return d.Truncate(time.Second).String()
 }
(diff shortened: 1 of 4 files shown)


pool.go
```diff
@@ -435,12 +432,7 @@ func (p *pool) fetchAndUpdate(ctx context.Context, node string, c cid.Cid, attem
 		}
 	}
 
-	var errTooManyRequests ErrSaturnTooManyRequests
-	if errors.As(err, &errTooManyRequests) {
-		err = &ErrSaturnTooManyRequests{
-			Node:       errTooManyRequests.Node,
-			RetryAfter: errTooManyRequests.RetryAfter,
-		}
```
hacdias, line 443: I have no idea why `err` was being replaced with an error of the same type with the same values. I just removed this.

caboose.go
```diff
@@ -102,27 +102,19 @@ var ErrContentProviderNotFound error = errors.New("saturn failed to find content
...
 type ErrSaturnTooManyRequests struct {
-	Node       string
-	RetryAfter time.Duration // TODO: DRY refactor after https://github.com/ipfs/go-libipfs/issues/188
+	Node string
 }
 
-func (e ErrSaturnTooManyRequests) Error() string {
-	return fmt.Sprintf("saturn node %s returned Too Many Requests error, please retry after %s", e.Node, humanRetry(e.RetryAfter))
+func (e *ErrSaturnTooManyRequests) Error() string {
+	return fmt.Sprintf("saturn node %s returned Too Many Requests error, please retry later", e.Node)
```
hacdias, line 109: Note that the actual value will be included on the wrapped error of type `gateway.ErrRetryAfter`.

caboose.go
```diff
@@ -102,27 +102,19 @@ var ErrContentProviderNotFound error = errors.New("saturn failed to find content
...
 
 func (e *ErrCidCoolDown) Error() string {
-	return fmt.Sprintf("multiple saturn retrieval failures seen for CID %s, please retry after %s", e.Cid, humanRetry(e.RetryAfter))
-}
-
-// TODO: move this to upstream error interface in https://github.com/ipfs/go-libipfs/issues/188
-// and refactor ErrCidCoolDown and ErrSaturnTooManyRequests to inherit from that instead
-func humanRetry(d time.Duration) string {
-	return d.Truncate(time.Second).String()
+	return fmt.Sprintf("multiple saturn retrieval failures seen for CID %s, please retry later", e.Cid)
```
hacdias, line 117: See previous comment.

failure_test.go
```diff
@@ -30,7 +32,9 @@ func TestHttp429(t *testing.T) {
 	_, err := ch.c.Get(ctx, testCid)
 	require.Error(t, err)
 
-	ferr := err.(*caboose.ErrSaturnTooManyRequests)
+	var ferr *gateway.ErrorRetryAfter
```
willscott, line 35: This is going to create a circular dependency.

the gateway uses caboose as it's blockstore / fetcher
us going back to the gateway for errors seems like the wrong direction

can we just define a common interface, e.g. if the error exposes a `RetryAfter()` method that we can conform to?

willscott (changes requested):

failure_test.go
hacdias: But caboose already requires `go-libipfs`: https://github.com/filecoin-saturn/caboose/blob/main/go.mod#L9

willscott: maybe put another way:
caboose is a common saturn client
why would it be returning gateway-specific errors?

lidel: I don't feel strongly. If we want to keep caboose consumer-agnostic, it is fine to close this PR, and keep explicit error translation in: https://github.com/ipfs/bifrost-gateway/pull/56/files/d65231f2119edb64a61d6044bd758181a068bb35#diff-4d0eb65b9891ffe23bd03c7596f2b573a5ecd5378e939ab7801acb5dca97b147R87

hacdias: It's also fine by me. We just have to remove the comments in the current caboose code that say the code will be refactored with the gateway errors to avoid future confusion.

lidel: @hacdias mind refactoring this PR to only remove these comments? :pray:

willscott: 🙏

i suspect as this becomes 'the way to access data from saturn with retries that work' we'll find that all the other clients also realize they also want the logic here as well :)

willscott (approved):

aarshkshah1992:
I am working on fixing the flaky tests. Please feel free to merge.

PR #51, opened 2023-02-25
support interface for non-block requests

willscott:
`fetchResource` is an underlying fetch request that can be called in a block or non-block context.

This isn't a complete abstraction, as the `fetchWith` logic will need to be also be applied on the in-development car interface.

This currently separates an individual block request / validation from the the internal request and tracking.

fix #45

PR #61, opened 2023-03-06
[WIP] Use observed speed in weighing

aarshkshah1992:
Give a weight boost to low latency nodes that serve fast fetches.  This will help us pick fast nodes for retrieval and create a natural latency based partition of L1s among the Bifrost nodes spread across the globe.


willscott:
consider doing this on top of #51 so we don't end up with another painful merge?

aarshkshah1992:
@willscott Are we confident of landing #51 anytime soon ? If yes, happy to do this but don't want to block this work on that PR.

willscott:
yeah, i'm happy to just merge it - it doesn't break your current code path, it just adds a new one

aarshkshah1992:
Okay, lemme review #51  and base this on top of that.

PR #62, opened 2023-03-06
Proposal for a 'per-request prefetch'

willscott:
This allows a sub-blockstore interface to be used at a request level specifying an initial expectation of the path -> blocks needed.

I suspect that this can initially be the inbound path to the gateway with '?depth=1' added in unixfs handler and '?depth=all' for tar/car handlers.

That will trigger loading of the requested car into a local cache in front of the blockstore, and blocks reading of the stream from the backing request to be reasonably in line with reads.

### DIFF
--- a/stoker/loosesync.go
+++ b/stoker/loosesync.go
@@ -1,0 +1,73 @@
+package stoker
+
+import (
+	"sync/atomic"
+	"time"
+)
+
+type writeAheadSynchronizer struct {
+	max     int
+	min     int
+	current atomic.Int32
+	notify  chan struct{}
+}
+
+func NewWAS(max, min int) *writeAheadSynchronizer {
+	return &writeAheadSynchronizer{
+		max:     max,
+		min:     min,
+		current: atomic.Int32{},
+		notify:  make(chan struct{}, 1),
+	}
+}
+
+// requst to add 1 to the queue, potentially blocking until that's cokay
+func (w *writeAheadSynchronizer) ReqAdd() {
+	n := w.current.Add(1)
+	if n >= int32(w.max) {
+		w.wait()
+	}
+}
+
+func (w *writeAheadSynchronizer) wait() {
+	done := false
+	t := time.NewTimer(15 * time.Millisecond)
+	for {
+		select {
+		case <-t.C:
+		case <-w.notify:
+		}
+
+		curr := w.current.Load()
+		if curr < int32(w.min) {
+			done = true
+		}
+
+		if !t.Stop() {
+			// Exhaust expired timer's chan.
+			select {
+			case <-t.C:
+			default:
+			}
+		}
+		if done {
+			return
+		}
+		t.Reset(15 * time.Millisecond)
+	}
+}
+
+// decrement 1 from the queue, potentially unblocking writers
+func (w *writeAheadSynchronizer) Dec() {
+	curr := w.current.Add(-1)
+	if curr < 0 {
+		// todo: okay that not atomic? i think we're okay leaking in this direction
+		w.current.Store(0)
+	}
+	if curr <= int32(w.min) {
+		select {
+		case w.notify <- struct{}{}:
+		default:
+		}
+	}
+}
--- a/stoker/stoker.go
+++ b/stoker/stoker.go
@@ -1,0 +1,145 @@
+// Stoker loads a dag from a remote host into an in-memory blockstore with the
+// intention that it will be processed in a traversal-order compatible way such
+// that data can be efficiently streamed and not fully buffered in memory.
+package stoker
+
+import (
+	"context"
+	"io"
+
+	"github.com/filecoin-saturn/caboose"
+	"github.com/ipld/go-car"
+
+	lru "github.com/hashicorp/golang-lru"
+	"github.com/ipfs/go-cid"
+	ipfsblockstore "github.com/ipfs/go-ipfs-blockstore"
+	"github.com/ipfs/go-libipfs/blocks"
+	golog "github.com/ipfs/go-log/v2"
+)
+
+var goLogger = golog.Logger("caboose/stoker")
+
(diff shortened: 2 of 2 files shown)


stoker/stoker.go
```diff
@@ -0,0 +1,145 @@
...
+func (ss *stokerSession) fillCache(ctx context.Context, path string) error {
+	err := ss.c.Fetch(ctx, path, func(resource string, reader io.Reader) error {
+		cr, err := car.NewCarReader(reader)
+		if err != nil {
+			return err
+		}
+
+		n := 0
+		for {
+			ss.writeAheadSynchronizer.ReqAdd()
```
aarshkshah1992, line 64: IIUC, this is to limit reading of incoming blocks from the CAR stream to keep memory usage in check, right ?
Given that the cache size is bounded, I don't think we need this to limit how many blocks we put in the cache.

aarshkshah1992: Basically, what would happen if we didn't use a `writeAheadSynchronizer ` here ?

stoker/stoker.go
```diff
@@ -0,0 +1,145 @@
...
+	blk, err := ss.getUncached(ctx, it)
+	return blk != nil, err
+}
+
+func (ss *stokerSession) getUncached(ctx context.Context, it cid.Cid) (blocks.Block, error) {
+	blk, err := ss.c.Get(ctx, it)
+	if err != nil {
+		return nil, err
+	}
+	ss.cache.Add(it, blk)
```
aarshkshah1992, line 98: Why do we NOT need to increment the count on the `writeAheadSynchronizer ` here ?

stoker/stoker.go
```diff
@@ -0,0 +1,145 @@
...
+	*caboose.Caboose
+}
+
+func New(c *caboose.Caboose) *Stoker {
+	return &Stoker{c}
+}
+
+const SessionCacheSize = 1024
+
+func (s *Stoker) NewSession(ctx context.Context, path string) ipfsblockstore.Blockstore {
```
aarshkshah1992, line 32: Will bifrost directly use the `Stoker` blockstore instead of the `Caboose` blockstore by giving it a path here ?

aarshkshah1992: @lidel Is it easy for Bifrost to pass the `path` here ?

stoker/stoker.go
```diff
@@ -0,0 +1,145 @@
...
+	return blk, nil
+}
+
+func (ss *stokerSession) Get(ctx context.Context, it cid.Cid) (blocks.Block, error) {
+	if cached, ok := ss.cache.Get(it); ok {
+		// todo: maybe we should only do this on reads against the cids coming
+		// from pre-fetching. that said, this will more likely read out the
+		// fill pre-fetch stream, which is still a reasonable choice.
+		ss.writeAheadSynchronizer.Dec()
+
```
aarshkshah1992, line 108: What does calling `Dec` here signify ? We're not really removing anything from the cache here. Is it like a "yeah, your prefetching is useful to me, keep doing what you are doing". Are we optimizing for memory usage here ?

stoker/loosesync.go
```diff
@@ -0,0 +1,73 @@
...
+			return
+		}
+		t.Reset(15 * time.Millisecond)
+	}
+}
+
+// decrement 1 from the queue, potentially unblocking writers
+func (w *writeAheadSynchronizer) Dec() {
+	curr := w.current.Add(-1)
+	if curr < 0 {
```
aarshkshah1992, line 63: This can go below zero because the same cached block can get read multiple times by Bifrost, right ? Do we want to decrement the count in this case where the same cached block gets read multiple times ?

stoker/stoker.go
aarshkshah1992: Also @lidel  -> Would it make sense to share the block cache across requests given IPFS GW access patterns ?

lidel: > Will bifrost directly use the Stoker blockstore instead of the Caboose blockstore by giving it a path here ?

We won't know until we finish [go-libipfs/gateway refactor](https://github.com/ipfs/go-libipfs/pull/176).
When done, bifrost-gateway may use it instead of blockstore for some types of requests, and you will get clear request for specific depth or byte ranges.

> Is it easy for Bifrost to pass the path here that it gets from it's clients ?

Caboose already uses  the requested Content Path as the Affinity key [here](https://github.com/filecoin-saturn/caboose/blob/e621c985ff5baf4f229ebd949b9916e2f8b8f235/caboose.go#L271-L281). 

This is the parent, normalized path (so you get uniform behavior across path, subdomain, and dnslink gateway requests based on `Host` HTTP header too). 

The problem is, you don't know how much you need to prefetch to satisfy client's request.
This PR is just guessing the most probable pattern (overfetching).

> When does passing the path to Lassie make more sense than passing the "affinity key" to Lassie ?

When you know the context (full file, only dir enumeration, full dag).
If you don't know the context, sending path to lassie will force preload of full dag (too much) or only root level (too little if request was for a file), and most likely impact earnings of L1s. 

> Would it make sense to share the block cache across requests given IPFS GW access patterns ?

We already have [global block cache in bifrost-gateway](https://github.com/ipfs/bifrost-gateway/blob/main/blockstore_cache.go) – 2Q cache is based on frequency+recency to match gateway usage patterns, and [has hit ratio of 40%-50%](https://github.com/ipfs/bifrost-gateway/issues/47). 

If you add any additional cache, add a metric which will let us see if it brings any value on top of the existing one.

lidel:
I believe we can close this since we now have GRAPH_BACKEND from https://github.com/ipfs/bifrost-gateway/pull/61.
Follow up work is tracked in https://github.com/ipfs/bifrost-gateway/issues/62 and https://github.com/ipfs/bifrost-gateway/issues/47

PR #68, opened 2023-03-29
feat: switch to boxo and fix CAR fetch timeouts

lidel:
This PR switches to `boxo` (why: https://github.com/ipfs/boxo/issues/215).

It is possible we will be also adding things related to https://github.com/ipfs/boxo/pull/176 and https://github.com/ipfs/bifrost-gateway/pull/61 to this PR, so everything can be reviewed and merged together.

(Usually trying to have smaller PRs, but we have too many PRs caused by `boxo` rename, trying to minimize overhead related to this unplanned rename)

## TODO

- [x] switch go.mod from `go-libipfs` to `boxo` (Context: https://github.com/ipfs/boxo/issues/215)
- [x]  go-car/v2 → boxo/ipld/car/v2 (Context: https://github.com/ipfs/boxo/issues/218#issuecomment-1486217454, avoiding double deps in bifrost-gateway)
- [x] use commit from https://github.com/ipfs/boxo/pull/176#issuecomment-1488202702 to unblock work on https://github.com/ipfs/bifrost-gateway/pull/61
- [x] fix timeouts (19s is way too low for CAR stream of unknown length)

## TBD, could be follow-up PR

- [ ] clean up Fetch API
- [x] verify timeouts 
- [ ] verify metrics


fetcher.go
```diff
@@ -145,7 +150,16 @@ func (p *pool) fetchResource(ctx context.Context, from string, resource string,
...
+	// TODO: Ideally, we would have additional "PerRequestInactivityTimeout"
+	// which is the amount of time without any NEW data from the server, but
+	// that can be added later. We need both because a slow trickle of data
+	// could take a large amount of time.
+	requestTimeout := DefaultSaturnCarRequestTimeout
+	if isBlockRequest {
+		requestTimeout = DefaultSaturnBlockRequestTimeout
+	}
+
+	reqCtx, cancel := context.WithTimeout(ctx, requestTimeout)
```
lidel, line 162: :thought_balloon: :warning:   we may have a DoS vector here, keeping meaningless connection open for 30 minutes is not the best.

Implementing "PerRequestInactivityTimeout"  would help a lot – we could then have this 30m timeout as a hard ceiling (or even raise it), but then have the same timeout for block and for CAR when  L1 did not send any new bytes for some time.

caboose.go
```diff
@@ -75,11 +75,19 @@ type Config struct {
 	MaxNCoolOff int
 }
 
+const DefaultLoggingInterval = 5 * time.Second
+const DefaultSaturnLoggerRequestTimeout = 1 * time.Minute
+
+const DefaultSaturnOrchestratorRequestTimeout = 30 * time.Second
+
+const DefaultSaturnBlockRequestTimeout = 19 * time.Second
+const DefaultSaturnCarRequestTimeout = 30 * time.Minute
```
lidel, line 84: :information_source: Had to add separate timeout for CARs. 
Not feeling strongly about 30m, but that is for how long I was able to stream wikipedia DAG from the old gateway, so a good starting point.

caboose.go
```diff
@@ -122,7 +130,16 @@ type ErrCoolDown struct {
 }
 
 func (e *ErrCoolDown) Error() string {
-	return fmt.Sprintf("multiple saturn retrieval failures seen for CID %s/Path %s, please retry after %s", e.Cid, e.Path, humanRetry(e.retryAfter))
```
lidel, line 125: :information_source:  This was producing very confusing log errors `failures seen for CID b/Path` due to `e.Cid` being undefined.

aarshkshah1992 (approved):

PR #69, opened 2023-03-30
forwards server timings up to context

willscott:
* add a timing metric for each caboose request around fetches when a timing key is present on the context
* forwards with a tag any timing keys in the response of the request onto the context timing key

### DIFF
--- a/fetcher.go
+++ b/fetcher.go
@@ -3,6 +3,7 @@ package caboose
 import (
 	"context"
 	"fmt"
+	"hash/crc32"
 	"io"
 	"net/http"
 	"strconv"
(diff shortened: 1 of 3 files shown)


willscott:
Example observed final metrics:
```
Server-Timing: fetch;dur=1657.943375,nginx;dur=72;subreq="b36e4244";attempt=0,nginx_uct;subreq="b36e4244";attempt=0,nginx_uht;subreq="b36e4244";attempt=0,nginx_urt;subreq="b36e4244";attempt=0
```

aarshkshah1992 (approved):

PR #70, opened 2023-03-31
Metrics Refactor and record metrics around L1 Server-Timings

aarshkshah1992:
Some refactor/fixes of the metrics code for correctness/grokability/dryness
Add more dimensions/labels to metrics to better capture block/car, cache-hit/cache-miss  cases
Record prometheus metrics around important L1 server-timings

### DIFF
--- a/fetcher.go
+++ b/fetcher.go
@@ -137,7 +137,7 @@ func (p *pool) fetchResource(ctx context.Context, from string, resource string, 
 				fetchDurationPerCarPerPeerSuccessMetric.WithLabelValues(cacheStatus).Observe(float64(response_success_end.Sub(start).Milliseconds()))
 			}
 
-			updateSuccessServerTimingMetrics(respHeader.Values(servertiming.HeaderKey), resourceType, cacheStatus)
+			updateSuccessServerTimingMetrics(respHeader.Values(servertiming.HeaderKey), resourceType, cacheStatus, durationMs, ttfbMs, received)
 		} else {
 			if isBlockRequest {
 				fetchDurationPerBlockPerPeerFailureMetric.Observe(float64(time.Since(start).Milliseconds()))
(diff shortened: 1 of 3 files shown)


metrics.go
```diff
@@ -35,101 +44,136 @@ var (
...
+		Help:    "Latency observed during failed caboose fetches from a single peer in milliseconds",
+		Buckets: durationMsPerBlockHistogram,
 	})
 
-	fetchDurationPerCarPerPeerFailureMetric = prometheus.NewHistogram(prometheus.HistogramOpts{
-		Name:    prometheus.BuildFQName("ipfs", "caboose", "fetch_duration_car_peer_failure"),
-		Help:    "Latency observed during failed caboose car fetches from a single peer",
-		Buckets: durationPerCarHistogram,
+	fetchDurationBlockFailureMetric = prometheus.NewHistogram(prometheus.HistogramOpts{
+		Name:    prometheus.BuildFQName("ipfs", "caboose", "fetch_duration_block_failure"),
```
willscott, line 106: renaming current metrics will break current dashboards - why is this changing?

aarshkshah1992: @willscott Not all the metrics have been renamed. Only some that were either incorrectly named or some that have taken the place of some repetitive ones.

I'll ensure the Caboose/Bifrost dashboard/metrics encompass the correct names. We're anyways revamping both the dashboards.

fetcher.go
```diff
@@ -284,6 +284,11 @@ func updateSuccessServerTimingMetrics(timingHeaders []string, resourceType strin
 					fetchDurationPerPeerSuccessCacheMissTotalLassieMetric.WithLabelValues(resourceType).Observe(float64(m.Duration.Milliseconds()))
 				case "nginx":
 					fetchDurationPerPeerSuccessTotalL1NodeMetric.WithLabelValues(resourceType, cache_status).Observe(float64(m.Duration.Milliseconds()))
+					networkTimeMs := totalTimeMs - m.Duration.Milliseconds()
+					fetchNetworkSpeedPerPeerSuccessMetric.WithLabelValues(resourceType).Observe(float64(recieved) / float64(networkTimeMs))
```
willscott, line 288: what about when `networkTimeMs` is 0? - that leads to a division by 0

willscott (approved):

fetcher.go
aarshkshah1992: Fixed and added more sanity checks.

PR #79, opened 2023-04-05
Instrument context failures and Instrument conn latency

aarshkshah1992:


fetcher.go
```diff
@@ -258,21 +271,38 @@ func (p *pool) fetchResource(ctx context.Context, from string, resource string,
...
 
 	// drain body so it can be re-used.
 	_, _ = io.Copy(io.Discard, resp.Body)
-
 	if err != nil {
 		return
 	}
 
 	response_success_end = time.Now()
+	{
```
willscott, line 286: why is this in a block?

willscott (approved):
remove the extraneous block before merge

fetcher.go
aarshkshah1992: removed

fetcher.go
```diff
@@ -71,6 +72,11 @@ func (p *pool) fetchResource(ctx context.Context, from string, resource string,
 	if mime == "application/vnd.ipld.raw" {
 		resourceType = "block"
 	}
+	fetchCalledTotalMetric.WithLabelValues(resourceType).Add(1)
+	if ctx.Err() != nil {
+		fetchRequestContextErrorTotalMetric.WithLabelValues(resourceType, ctx.Err().Error(), "init").Add(1)
```
lidel, line 77: :warning:   @aarshkshah1992 @willscott  this is potentially really really bad, because you have unbounded number of values `.Error()`   could return, which will in turn baloon the size of metrics.

Last time I checked we had errors which had the URL / CID of failed request, and that means every error will be unique, exploding metrics space:


```
http request failed: Get \"https://64.44.166.190/ipfs/bafybeiaysi4s6lnjev27ln5icwm6tueaw2vdykrtjkwiphwekaywqhcjze/favicon.ico?format=car&depth=0\": context canceled
```

willscott: https://github.com/filecoin-saturn/caboose/pull/81

aarshkshah1992: @lidel Thanks for catching this ! In my mind, the context cancelled error could only be Cancelled or Dead line exceeded ! But I see now.

PR #80, opened 2023-04-05
Ensure deadline errors are legit

aarshkshah1992:


aarshkshah1992:
@willscott Agreed. I think the whole metrics collection code is going to need a refactor but resisting the temptation to do so right now while we are focused on bigger fish.

willscott:
do we see a count of errors  from this fetchIncorrectDeadlineErrorTotalMetric?
if not, do we want to revert this to reduce the complexity?

PR #85, opened 2023-04-07
More instrumentation of context cancellations and reward low latency nodes

aarshkshah1992:
1. We do have some peers with a reasonable P75 latency  even after they have served >100 successful cache hit retrievals
https://protocollabs.grafana.net/d/6g0_YjBVk/bifrost-caboose-staging?orgId=1&from=now-6h&to=now&editPanel=32

2. Even after the context cancellation fix by @willscott in https://github.com/filecoin-saturn/caboose/pull/82, a significant number of requests(for both blocks and CARs) still fail with the context cancelled error.  Let's instrument them more to help Bifrost debug this.
https://protocollabs.grafana.net/d/6g0_YjBVk/bifrost-caboose-staging?orgId=1&from=now-6h&to=now&editPanel=30

### DIFF
--- a/caboose.go
+++ b/caboose.go
@@ -256,6 +256,10 @@ func (c *Caboose) Close() {
 
 // Fetch allows fetching car archives by a path of the form `/ipfs/<cid>[/path/to/file]`
 func (c *Caboose) Fetch(ctx context.Context, path string, cb DataCallback) error {
+	fetchCalledTotalMetric.WithLabelValues(resourceTypeCar).Add(1)
+	if recordIfContextErr(resourceTypeCar, ctx, "FetchApi") {
+		return ctx.Err()
+	}
 	return c.pool.fetchResourceWith(ctx, path, cb, c.getAffinity(ctx))
 }
 
@@ -268,6 +272,10 @@ func (c *Caboose) Has(ctx context.Context, it cid.Cid) (bool, error) {
 }
 
 func (c *Caboose) Get(ctx context.Context, it cid.Cid) (blocks.Block, error) {
+	fetchCalledTotalMetric.WithLabelValues(resourceTypeBlock).Add(1)
+	if recordIfContextErr(resourceTypeBlock, ctx, "FetchBlockApi") {
+		return nil, ctx.Err()
+	}
 	blk, err := c.pool.fetchBlockWith(ctx, it, c.getAffinity(ctx))
 	if err != nil {
 		return nil, err
--- a/caboose.go
+++ b/caboose.go
@@ -256,10 +256,6 @@ func (c *Caboose) Close() {
 
 // Fetch allows fetching car archives by a path of the form `/ipfs/<cid>[/path/to/file]`
 func (c *Caboose) Fetch(ctx context.Context, path string, cb DataCallback) error {
-	fetchCalledTotalMetric.WithLabelValues(resourceTypeCar).Add(1)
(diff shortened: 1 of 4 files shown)


caboose.go
```diff
@@ -268,6 +272,10 @@ func (c *Caboose) Has(ctx context.Context, it cid.Cid) (bool, error) {
 }
 
 func (c *Caboose) Get(ctx context.Context, it cid.Cid) (blocks.Block, error) {
+	fetchCalledTotalMetric.WithLabelValues(resourceTypeBlock).Add(1)
+	if recordIfContextErr(resourceTypeBlock, ctx, "FetchBlockApi") {
+		return nil, ctx.Err()
+	}
```
willscott, line 278: this is only a partial view of this, right? - because asks for `Has` and `GetSize` also trigger parallel block triggers to `Get`?

pool.go
```diff
@@ -245,6 +239,20 @@ func (p *pool) doRefresh() {
...
 		}
 
+		// give weight bumps to low latency peers that have served > 100 successful "low latency" cache hit retrievals.
+		poolWeightBumpMetric.Set(0)
+		for _, m := range n {
+			m := m
+			if perf, ok := p.nodePerf[m.url]; ok {
+				// Our analysis so far shows that we do have ~10-15 peers with -75 < 200ms latency.
+				// It's not the best but it's a good start and we can tune as we go along.
+				if perf.latencyDigest.Count() > 100 && perf.latencyDigest.Quantile(0.75) <= 200 {
```
willscott, line 249: these parameters probably shouldn't be hard coded? - maybe environment variables at least?

caboose.go
aarshkshah1992: @willscott Yeah, just wanna focus on fetch for now.

willscott: then lets not record the block api at all - misreporting will lead to confusion?

pool.go
aarshkshah1992: @willscott 

Fixed the instrumentation. We now record on the lower level block and car fetchBlockWith and fetchResourceWith instead of the public API for capturing accurate numbers.

caboose.go
aarshkshah1992: @willscott 

Fixed the instrumentation. We now record on the lower level block and car fetchBlockWith and fetchResourceWith instead of the public API for capturing accurate numbers.

pool.go
aarshkshah1992: @willscott Made these varibales for now. Will make them env in the next Caboose PR that I'm gonna write for much better pool membership and weighing algo.

metrics.go
```diff
@@ -58,8 +58,8 @@ var (
 	durationMsPerCarHistogram = prometheus.ExponentialBucketsRange(50, 1800000, 40)
 
 	// buckets to measure latency between a caboose peer a Saturn L1,
-	// [50ms, 100ms, 200ms, ...,  ~25 seconds]
-	latencyDistMsHistogram = prometheus.ExponentialBuckets(50, 2, 10)
+	// [50ms, 75ms, 100ms, ...,  525 ms]
+	latencyDistMsHistogram = prometheus.LinearBuckets(50, 25, 20)
```
willscott, line 62: can we start at 0 (add the 0-25,25-50?)

willscott (approved):

metrics.go
aarshkshah1992: This is done.

PR #86, opened 2023-04-12
Implement Tiered Hashing with small pool sizes, kick out nodes for correctness, talk more to fast nodes and reduce pool churn

aarshkshah1992:
### TODO

- [x] Run in staging  → https://github.com/ipfs/bifrost-gateway/pull/94
- [x] Unit Tests
- [x] Green CI
- [x] Investigate and fix low cache hit Saturn pcts 

**_Subsequent PRs:_**
- [ ] Traffic Mirroring to probe the unknown set instead of the current  probabilistic mechanism ? (subsequent PR)
-  [ ] Use exponential decay over the entire history of the node rather than sliding window percentiles to rank nodes


caboose.go
```diff
@@ -16,6 +16,11 @@ import (
 	"github.com/ipfs/go-cid"
 )
 
+const (
+	BifrostProd    = "bifrost-prod"
+	BifrostStaging = "bifrost-staging"
+)
+
```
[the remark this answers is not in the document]
aarshkshah1992: @willscott Yeah, had this for local testing. Will move to using env vars once we set them on Bifrost.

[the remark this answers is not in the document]

willscott (approved):
This seems fine for a next iteration

The "speed" calculation I think needs to look at bytes received versus wall clock time, rather than the throughput of individual requests

aarshkshah1992:
@willscott Not sure what you mean by that speed comment. Speed is calculated as (bytes recieved)/(end - start time).

What change do we need to make here ?

willscott:
That the speed we care about is how many bytes we got over the last minute - we'll get more accuracy by adding receiver bytes into a bucket per peer and looking at it overall, rather than sampling each individual request as individual throughput measurements

aarshkshah1992:
@willscott But we add the speed of the individual request to a bucket for that peer and then use the P25 of that bucket to rank. Isn't that what you are saying here ?

fetcher.go
```diff
@@ -187,7 +187,7 @@ func (p *pool) fetchResource(ctx context.Context, from string, resource string,
 				// my address
 				Range:          "",
 				Referrer:       respReq.Referer(),
-				UserAgent:      respReq.UserAgent(),
+				UserAgent:      "bifrost-" + os.Getenv(EnvironmentKey),
```
willscott, line 190: [the embedding bifrost-gatway](https://github.com/ipfs/bifrost-gateway/blob/3c65632495c8cc6e0db55103c0d9deea918204bb/http_client.go#L17) is already setting a useful version in the user agent. we shouldn't drop that.

aarshkshah1992: @willscott Okay, will append

failure_test.go
```diff
@@ -260,6 +188,7 @@ func (e *ep) Setup() {
 	e.valid = true
 	e.resp = testBlock
 	e.server = httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		time.Sleep(time.Millisecond * 20)
```
willscott, line 191: why is this needed? / why are responses delayed?

fetcher.go
```diff
@@ -74,14 +75,17 @@ func (p *pool) doFetch(ctx context.Context, from string, c cid.Cid, attempt int)
 }
 
 // TODO Refactor to use a metrics collector that separates the collection of metrics from the actual fetching
-func (p *pool) fetchResource(ctx context.Context, from string, resource string, mime string, attempt int, cb DataCallback) (err error) {
+// rm will be nil only for context cancellation errors
```
willscott, line 78: this is confusing - it looks like it isn't nil even in the context cancellation case - it's just an empty ResponseMetrics objects?

fetcher.go
```diff
@@ -232,22 +227,35 @@ func (p *pool) fetchResource(ctx context.Context, from string, resource string,
 			}
 		}
 	}
+	req.Header.Add("User-Agent", "bifrost-"+os.Getenv(EnvironmentKey))
```
willscott, line 230: don't over-ride the provided user agent - there's already a more specific user agent set on the client that bifrost passes in and this will over-write it.

out.car
willscott, line 1: this shouldn't be committed / in history

metrics.go
```diff
@@ -236,13 +207,34 @@ var (
...
 	CabooseMetrics.MustRegister(poolSizeMetric)
-	CabooseMetrics.MustRegister(poolHealthMetric)
 	CabooseMetrics.MustRegister(poolNewMembersMetric)
+	CabooseMetrics.MustRegister(poolRemovedFailureTotalMetric)
+	CabooseMetrics.MustRegister(poolRemovedConnFailureTotalMetric)
+	CabooseMetrics.MustRegister(poolRemovedReadFailureTotalMetric)
+	CabooseMetrics.MustRegister(poolRemovedNon2xxTotalMetric)
+	CabooseMetrics.MustRegister(poolMembersNotAddedBecauseRemovedMetric)
+	CabooseMetrics.MustRegister(poolEnoughObservationsForMainSetDurationMetric)
+	CabooseMetrics.MustRegister(poolTierChangMetric)
```
willscott, line 237: `poolTierChangeMetric`

pool.go
```diff
@@ -6,27 +6,23 @@ import (
...
+
 	blocks "github.com/ipfs/go-block-format"
 	"github.com/ipfs/go-cid"
 	"github.com/patrickmn/go-cache"
-	"github.com/serialx/hashring"
 )
 
-const maxPoolSize = 300
-
 var (
```
willscott, line 23: const

pool.go
```diff
@@ -172,132 +87,53 @@ func (p *pool) Start() {
...
-				// It's not the best but it's a good start and we can tune as we go along.
-				if perf.latencyDigest.Count() > float64(minSuccessfulRetrievals) && perf.latencyDigest.Quantile(latencyPercentile) <= maxLatency {
-					poolWeightBumpMetric.Add(1)
-					m.weight = maxWeight
-				}
-			}
-		}
+	mt := p.th.GetPoolMetrics()
+	poolSizeMetric.WithLabelValues("unknown").Set(float64(mt.Unknown))
+	poolSizeMetric.WithLabelValues("main").Set(float64(mt.Main))
```
willscott, line 115: "unknown" / "main" also as consts

pool.go
```diff
@@ -362,9 +198,16 @@ func (p *pool) fetchBlockWith(ctx context.Context, c cid.Cid, with string) (blk
...
 
-	nodes, err := p.getNodesToFetch(cidToKey(c), with)
-	if err != nil {
-		return nil, err
+	aff := with
+	if aff == "" {
+		aff = cidToKey(c)
+	}
+
+	p.lk.Lock()
```
willscott, line 206: why is this an exclusive lock?

pool.go
```diff
@@ -509,9 +285,16 @@ func (p *pool) fetchResourceWith(ctx context.Context, path string, cb DataCallba
...
 
-	nodes, err := p.getNodesToFetch(path, with)
-	if err != nil {
-		return err
+	aff := with
+	if aff == "" {
+		aff = path
+	}
+
+	p.lk.Lock()
```
willscott, line 293: likewise: why is this an exclusive lock?

tieredhashing/sorting.go
```diff
@@ -0,0 +1,42 @@
...
+	"sort"
+
+	"github.com/asecurityteam/rolling"
+)
+
+func (t *TieredHashing) nodesSortedLatency() []nodeWithLatency {
+	var nodes []nodeWithLatency
+
+	for n, perf := range t.nodes {
+		perf := perf
```
willscott, line 13: ?

tieredhashing/tiered_hashing.go
```diff
@@ -0,0 +1,384 @@
...
+	"time"
+
+	"github.com/asecurityteam/rolling"
+
+	"github.com/patrickmn/go-cache"
+
+	"github.com/serialx/hashring"
+)
+
+// TODO Make env vars for tuning
```
willscott, line 15: file an issue for this if it's not in the PR

failure_test.go
aarshkshah1992: Just want to introduce some form of delay in the tests as this is the "L1 peer that sends us blocks".

tieredhashing/tiered_hashing.go
aarshkshah1992: https://github.com/filecoin-saturn/caboose/pull/86

fetcher.go
aarshkshah1992: @willscott Can we append this to the existing `UserAgent` that the bifrost client has on it ? I believe L1s need this to be able to bifurcate metrics etc on their side based on environment.

caboose_test.go
```diff
@@ -205,3 +189,79 @@ func TestResource(t *testing.T) {
 		t.Fatal("expected fall-over progress")
 	}
 }
+
+type HarnessOption func(config *caboose.Config)
+
+func WithTieredHashingOpts(opts []tieredhashing.Option) func(config *caboose.Config) {
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

tieredhashing/tiered_hashing.go
aarshkshah1992: Done.

pool.go
aarshkshah1992: Yeah agreed this can be just a read lock. Fixed.

fetcher.go
aarshkshah1992: Yeah this is a remanant of a previous iteration. Fixed.

pool.go
aarshkshah1992: Done.

pool.go
aarshkshah1992: Done.

metrics.go
aarshkshah1992: Done.

out.car
aarshkshah1992: Removed.

fetcher.go
willscott: appending is okay, sure

willscott: or you could set environment as another header, right?

aarshkshah1992: Appended to existing

failure_test.go
willscott: doing so masks race conditions / makes tests artificially slow - we ideally shouldn't need it - add TODO to further clean up tests

willscott (approved):

PR #87, opened 2023-04-12
feat: open telemetry tracing

hacdias:
Adds Open Telemetry tracing spans. Context: https://github.com/ipfs/bifrost-gateway/issues/68#issuecomment-1504921191.

This is not necessarily required. On the screenshot below you see the tracing result from the Bifrost Gateway up until the HTTP request Caboose makes to the Saturn node. What this PR adds is the `Caboose.*` trace spans. The HTTP GET already worked.

Screenshot 2023-04-12 at 11 19 18

@lidel @willscott not sure where is the most value to put the spans here, but this seems like a good start. Feel free to add/remove some.

### DIFF
--- a/caboose.go
+++ b/caboose.go
@@ -14,6 +14,8 @@ import (
 	gateway "github.com/ipfs/boxo/gateway"
 	blocks "github.com/ipfs/go-block-format"
 	"github.com/ipfs/go-cid"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 )
 
 type Config struct {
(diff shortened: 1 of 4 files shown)


lidel (changes requested):
We should  remove `Blockstore.` prefix, at least for `Fetch`, which  is an Exchange interface, and not Blockstore one.

lidel (approved):

lidel (approved):
lgtm, should be good enough for now 
@willscott  / @aarshkshah1992  ok to merge? we can add more spans in the future, if needed

willscott (approved):

PR #90, opened 2023-05-02
Use mirroring of a fraction of traffic for testing unknown nodes

willscott:
sends a specified fraction of requests to a centrally managed sample queue that tests unknown tier nodes to track their latency


aarshkshah1992:
@willscott Let's hold off on this till we've finished our load testing with L1s as this PR will only really increase traffic on the main set of nodes. I am happy to own this and take this to completion.

aarshkshah1992:
@willscott I am NOT saying we don't need this man. Ofcourse we need this to improve higher percentile latencies.

 I am saying let's land/deploy this once we've figured out why a bunch of main nodes get kicked out for correctness/what kind of load they can handle. This PR will only really increase the load on the main set nodes without us first understanding what's going on wrt to current observed perf.

willscott:
Long lived branches with rebasing and merge conflicts seem more painful than rapid merge +iterate

aarshkshah1992:
@willscott 

1. I am happy to own the landing of this PR and take over from you. 
2. However, I don't want to see even more main nodes getting kicked out because we're not sending 25% more traffic to them than before without first understanding what's going on.
3. Is there a middle-ground where we can land this PR without turning this feature on ? I think that can be something that can be merged and we can "turn it on" later.

tieredhashing/tiered_hashing.go
```diff
@@ -199,43 +200,40 @@ func (t *TieredHashing) GetPoolMetrics() PoolMetrics {
...
-		nodes2, _ := t.unknownSet.GetNodes(key, t.unknownPossible(n-len(nodes)))
-		nodes = append(nodes, nodes2...)
 	}
 
 	return nodes
 }
 
+func (t *TieredHashing) NodeTier(node string) Tier {
+	for k, n := range t.nodes {
+		if k == node {
```
aarshkshah1992, line 230: Let's copy `n` which is a pointer var before using it ?

tieredhashing/tiered_hashing.go
```diff
@@ -199,43 +200,40 @@ func (t *TieredHashing) GetPoolMetrics() PoolMetrics {
...
-		if t.mainSet.Size() != 0 {
-			nodes, ok = t.mainSet.GetNodes(key, t.mainPossible(n))
-			if !ok {
-				return nil
-			}
-			if len(nodes) == n {
-				return nodes
-			}
+		if len(nodes) == n {
+			return nodes
```
aarshkshah1992, line 213: You can just return nodes here. Why check `if len(nodes) == n` ?

tieredhashing/tiered_hashing.go
```diff
@@ -199,43 +200,40 @@ func (t *TieredHashing) GetPoolMetrics() PoolMetrics {
...
+		}
+	} else if from == TierMain {
+		nodes, ok = t.mainSet.GetNodes(key, t.mainPossible(n))
+		if !ok {
+			return nil
+		}
+		if len(nodes) == n {
+			return nodes
```
aarshkshah1992, line 221: Why do we do this check ?

pool.go
```diff
@@ -69,10 +79,11 @@ type pool struct {
...
 	p := pool{
-		config:  c,
-		started: make(chan struct{}),
-		refresh: make(chan struct{}, 1),
-		done:    make(chan struct{}, 1),
+		config:        c,
+		started:       make(chan struct{}),
+		refresh:       make(chan struct{}, 1),
+		done:          make(chan struct{}, 1),
+		mirrorSamples: make(chan poolRequest, 10),
```
aarshkshah1992, line 86: why this magic number ?

pool.go
```diff
@@ -217,6 +298,15 @@ func (p *pool) fetchBlockWith(ctx context.Context, c cid.Cid, with string) (blk
 		if recordIfContextErr(resourceTypeBlock, ctx, "fetchBlockWithLoop") {
 			return nil, ctx.Err()
 		}
+
+		// sample request for mirroring
+		if p.config.MirrorFraction > rand.Float64() {
```
aarshkshah1992, line 303: We need a default value for `MirrorFraction`

pool.go
```diff
@@ -217,6 +298,15 @@ func (p *pool) fetchBlockWith(ctx context.Context, c cid.Cid, with string) (blk
...
+			select {
+			case p.mirrorSamples <- poolRequest{node: nodes[i], path: fmt.Sprintf("/ipfs/%s?format=raw", c), key: aff}:
```
aarshkshah1992, line 305: So we're converting a block request to a CAR request ? How does this work ? Will Nginx/L1 shim/Lassie just send back that one block in a CAR file ?

pool.go
```diff
@@ -166,6 +178,70 @@ func (p *pool) refreshPool() {
...
 }
 
+func (p *pool) checkPool() {
+	for {
+		select {
+		case msg := <-p.mirrorSamples:
+			// see if it is to a main-tier node - if so find appropriate test node to test against.
+			p.lk.RLock()
+			if p.th.NodeTier(msg.node) != tieredhashing.TierMain {
+				p.lk.RUnlock()
```
aarshkshah1992, line 188: Why do we need to see if this is a `TierMain ` node ? Will an unknown node request make it to the loop ? Oh if we don't have enough main nodes for the mainset requests ?

pool.go
```diff
@@ -166,6 +178,70 @@ func (p *pool) refreshPool() {
 	}
 }
 
+func (p *pool) checkPool() {
```
aarshkshah1992, line 181: Please can we add a unit test for check pool for block reqs and car reqs and for main nodes vs unknown nodes ?

pool.go
```diff
@@ -166,6 +178,70 @@ func (p *pool) refreshPool() {
...
+			trialTimeout, cancel := context.WithTimeout(context.Background(), time.Second*19)
+			_ = p.fetchResourceAndUpdate(trialTimeout, testNodes[0], msg.path, 0, p.mirrorValidator)
+			cancel()
+		case <-p.done:
+			return
+		}
+	}
+}
+
+func (p *pool) mirrorValidator(resource string, reader io.Reader) error {
```
aarshkshah1992, line 205: Please can we add a unit tests for `mirrorValidator ` happy and failure cases ?

aarshkshah1992:
@willscott Please can we add a top level "integration tests" in caboose_test.go or somewhere to test the mirroring e2e for block and car requests and make sure that stats get recorded in the tiered pool for the mirroed traffic as well ?

pool.go
willscott: it gives some slack for parallelism / how much of a backlog of requests are we letting build up - should be more than 1 probably, but shouldn't be huge

pool.go
willscott: 1% ?

pool.go
willscott: since we ask for `format=raw` we get back directly the block data, not even a car here - i that's a good catch - yes the intention here was to get a car with the single block back

pool.go
willscott: yeah, if we're already using unknown nodes because we don't have a good main pool then we don't need to also be doing mirror-based sampling.

pool.go
aarshkshah1992: Sure, let's start with that !

pool.go
aarshkshah1992: I meant if the node is already an unknown node, don't do mirroring vs do mirroring if the node is a main node.

tieredhashing/tiered_hashing.go
willscott: we're returning a direct string, so nothing about n is retained to get messed up outside of the loop

pool.go
willscott: i'm tempted to not put too much work into this because we should get an actual validator in both boxo and lassie and should switch to one of those two once they exist

pool.go
```diff
@@ -204,7 +205,12 @@ func (p *pool) checkPool() {
 
 // TODO: this should be replaced with a real validator once one exists from boxo.
 func (p *pool) mirrorValidator(resource string, reader io.Reader) error {
-	parse, err := path.ParsePath(resource)
+	// first get the 'path' part to remove query string if present.
+	pth, err := url.Parse(resource)
```
aarshkshah1992, line 223: @willscott What if the original request is a CAR request and NOT a block request ? Will this work ?

willscott: yeah, it shouldn't matter- all we're doing in this step and below is extracting the first cid that appears in the path. then we read whatever we get back as a car and make sure the car has that first CID

aarshkshah1992 (approved):
lgtm for now. I will add tests later and I'll add the next iteration of Caboose improvements on top of this.

PR #93, opened 2023-05-04
Prevent pool starvation during duress and record CAR response size vs TTFB

aarshkshah1992:
- Add back the nodes we've kicked out if they're the only option we have to fill the pool. However, do it in descending order of reputation as determined by the orchestrator.
-  Let's plot CAR response size vs TTFB to visualise the co-relation there if any.
- Let's use a `CLIENT_ENV` header instead of `User-Agent` for debugging/metrics on L1s.

### DIFF
--- a/caboose.go
+++ b/caboose.go
@@ -20,7 +20,8 @@ import (
 )
 
 const (
-	EnvironmentKey = "STRN_ENV_TAG"
+	STRN_ENV_KEY   = "STRN_ENV_TAG"
+	CLIENT_ENV_KEY = "CLIENT_ENV"
 )
 
 type Config struct {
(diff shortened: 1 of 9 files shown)


caboose.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 const (
-	EnvironmentKey = "STRN_ENV_TAG"
+	STRN_ENV_KEY = "STRN_ENV_TAG"
```
willscott, line 23: use `SaturnEnvKey` as this camel-case style is used for all other caps

willscott (approved):

caboose.go

PR #97, opened 2023-05-08
No-penalise-403s

aarshkshah1992:
These are simply forbidden because of bad bits.


willscott (approved):

willscott:
this can't be totally true

what if we get a 403 from a malicious L1, when the content isn't really a 'bad bits'?

aarshkshah1992:
@willscott For sure, but that is true for 502s/504s etc etc any status code you can think of. 

However, in the next iteration of Caboose, I am gonna put in code to penalise("less traffic and less monies for you") nodes that haven't sent a successful response in a while so take care of these things at an aggregate level.

PR #101, opened 2023-05-18
[WIP: Draft] Caboose prod changes

aarshkshah1992:


caboose.go
```diff
@@ -80,7 +80,7 @@ const DefaultSaturnBlockRequestTimeout = 19 * time.Second
 const DefaultSaturnCarRequestTimeout = 30 * time.Minute
 
 const DefaultMaxRetries = 3
-const DefaultMirrorFraction = 0.05
+const DefaultMirrorFraction = 0.5
```
willscott, line 83: this seems really high / why does this make sense as a default?

fetcher.go
```diff
@@ -261,17 +269,23 @@ func (p *pool) fetchResource(ctx context.Context, from string, resource string,
...
 	saturnCallsTotalMetric.WithLabelValues(resourceType).Add(1)
 	startReq := time.Now()
-	resp, err = p.config.SaturnClient.Do(req)
+
+	for i := 0; i < 3; i++ {
+		resp, err = p.config.SaturnClient.Do(req)
+		if err == nil {
+			break
+		}
+	}
```
willscott, line 278: this isn't something we want to do in prod, right?

fetcher.go
```diff
@@ -303,6 +318,9 @@ func (p *pool) fetchResource(ctx context.Context, from string, resource string,
 	}
 
 	if resp.StatusCode != http.StatusOK {
+		// empty body so it can be re-used.
+		_, _ = io.Copy(io.Discard, resp.Body)
```
willscott, line 322: this is is a great bug to fix! maybe we quickly promote just this change and see how much it fixes?

fetcher.go
aarshkshah1992: @willscott Why not ? What's wrong in retrying connections ? We're not increasing the timeout here and nodes that keep failing this will eventually get kicked out.

willscott: because we're already in a loop doing retries one level up, where when we retry we retry to an alternate node.
adding a second loop here means we're now amplifying requests 9x instead of 3x without it being part of the parameter for how much we retry that's able to be controlled

PR #102, opened 2023-05-19
Prod iteration 2 

aarshkshah1992:
- Removes tiers. I've repeatedly seen that we're getting **_way_** better perf and cache affinity and way less churn without it. We still haven't figured out what's causing those connection issues that causes nodes to get kicked out from the main set.  

https://protocollabs.grafana.net/d/6g0_YjBVk/bifrost-caboose-staging?orgId=1&from=now-1h&to=now&editPanel=83
https://protocollabs.grafana.net/d/6g0_YjBVk/bifrost-caboose-staging?orgId=1&from=now-5m&to=now&editPanel=18

- Introduces metrics around response sizes for failed CAR fetches and a demarcation between connection failures because of timeouts vs non-timeouts

https://protocollabs.grafana.net/d/6g0_YjBVk/bifrost-caboose-staging?orgId=1&editPanel=84

- Bifrost will now move to HTTP1.

We're seeing way better perf across the board now with the staging instance serving 20% of the entire Gateway traffic.

image

Let's see what prod turns up.

### DIFF
--- a/caboose.go
+++ b/caboose.go
@@ -80,7 +80,7 @@ const DefaultSaturnBlockRequestTimeout = 19 * time.Second
 const DefaultSaturnCarRequestTimeout = 30 * time.Minute
 
 const DefaultMaxRetries = 3
-const DefaultMirrorFraction = 0.2
+const DefaultMirrorFraction = 0.1
 
 const maxBlockSize = 4194305 // 4 Mib + 1 byte
 const DefaultOrchestratorEndpoint = "https://orchestrator.strn.pl/nodes/nearby?count=200"
--- a/caboose.go
+++ b/caboose.go
@@ -80,7 +80,7 @@ const DefaultSaturnBlockRequestTimeout = 19 * time.Second
 const DefaultSaturnCarRequestTimeout = 30 * time.Minute
 
 const DefaultMaxRetries = 3
-const DefaultMirrorFraction = 0.05
+const DefaultMirrorFraction = 0.2
 
 const maxBlockSize = 4194305 // 4 Mib + 1 byte
 const DefaultOrchestratorEndpoint = "https://orchestrator.strn.pl/nodes/nearby?count=200"
--- a/fetcher.go
+++ b/fetcher.go
@@ -262,27 +262,32 @@ func (p *pool) fetchResource(ctx context.Context, from string, resource string, 
 	var resp *http.Response
 	saturnCallsTotalMetric.WithLabelValues(resourceType).Add(1)
 	startReq := time.Now()
-	resp, err = p.config.SaturnClient.Do(req)
-	if err != nil {
-		// just retry-once
-		resp, err = p.config.SaturnClient.Do(req)
-		if err != nil {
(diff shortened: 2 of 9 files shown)


fetcher.go
```diff
@@ -261,7 +262,14 @@ func (p *pool) fetchResource(ctx context.Context, from string, resource string,
 	var resp *http.Response
 	saturnCallsTotalMetric.WithLabelValues(resourceType).Add(1)
 	startReq := time.Now()
-	resp, err = p.config.SaturnClient.Do(req)
+
+	count := 0
+	for ; count < 3; count++ {
```
aarshkshah1992, line 267: This does help. Proof here:

https://protocollabs.grafana.net/d/6g0_YjBVk/bifrost-caboose-staging?orgId=1&from=now-1h&to=now&editPanel=83

See no impact on latency after deploying this. Let's keep this on.

willscott: But it's totally unaccountable.
we can't just block on trying 3 times on one node as part of "one attempt"

aarshkshah1992: addressed this.

willscott (commented):
test in staging, confirm that the churn metric stays reasonable, and then seems good to promote like this to prod

willscott (approved):

aarshkshah1992:
Note that this has not yet been deployed to prod as I want to test this thoroughly.

caboose.go
```diff
@@ -80,7 +80,7 @@ const DefaultSaturnBlockRequestTimeout = 19 * time.Second
 const DefaultSaturnCarRequestTimeout = 30 * time.Minute
 
 const DefaultMaxRetries = 3
-const DefaultMirrorFraction = 0.05
+const DefaultMirrorFraction = 0.2
```
willscott, line 83: this is way too high

pool.go
```diff
@@ -489,6 +490,13 @@ func (p *pool) fetchBlockAndUpdate(ctx context.Context, node string, c cid.Cid,
 
 func (p *pool) fetchResourceAndUpdate(ctx context.Context, node string, path string, attempt int, cb DataCallback) (err error) {
 	rm, err := p.fetchResource(ctx, node, path, "application/vnd.ipld.car", attempt, cb)
+	if err != nil && rm.ConnFailure {
+		rm, err = p.fetchResource(ctx, node, path, "application/vnd.ipld.car", attempt, cb)
+		if err == nil {
+			saturnConnectionRetriesTotalMetric.WithLabelValues(resourceTypeCar).Add(1)
+		}
+	}
```
willscott, line 498: I continue to think this is not useful.

We have `attempt` already and are doing this in a loop. having one `attempt` involve sometimes blocking on a single node twice is counter-intuitive and hard to reason about.

caboose.go
aarshkshah1992: Reduced to 10%

pool.go
aarshkshah1992: Removed this.

willscott (approved):

PR #109, opened 2023-06-06
Instrument response read failures

aarshkshah1992:


caboose.go
```diff
@@ -206,7 +207,13 @@ func NewCaboose(config *Config) (*Caboose, error) {
...
 	if c.config.OrchestratorEndpoint == nil {
 		var err error
-		c.config.OrchestratorEndpoint, err = url.Parse(DefaultOrchestratorEndpoint)
+
+		u := DefaultOrchestratorEndpoint
+		if v := os.Getenv(SaturnOrchUrlEnvKey); len(v) > 0 {
+			u = DefaultOrchestratorCoreEndpoint
+		}
+
+		c.config.OrchestratorEndpoint, err = url.Parse(u)
```
willscott, line 216: why are we adding additional complexity here? we already have an env variable to set `OrchestratorEndpoint` - why wouldn't we pass in the custom 'core only' orchestrator endpoint that way ,versus this pair of over-rides?

in practice, orchestrator is, i think, set in bifrost-gateway which woudl mean neither of these over-rides would take effect?

fetcher.go
```diff
@@ -86,6 +87,11 @@ func (p *pool) doFetch(ctx context.Context, from string, c cid.Cid, attempt int)
 
 // TODO Refactor to use a metrics collector that separates the collection of metrics from the actual fetching
 func (p *pool) fetchResource(ctx context.Context, from string, resource string, mime string, attempt int, cb DataCallback) (rm tieredhashing.ResponseMetrics, err error) {
+	isRange := "no"
+	if strings.Contains(resource, "bytes") {
```
willscott, line 91: * should we parse the URL and make sure there's a query parameter with the name?
* the name should be `entity-bytes` not bytes, right?

caboose.go
aarshkshah1992: Oh right. I agree. I'll fix this.

PR #113, opened 2023-06-15
sentinel cid implementation

AmeanAsad:
## Description:
This allows caboose to fetch sentinel cids from the orchestrator using authenticated requests and randomly query those cids.


pool.go
```diff
@@ -28,26 +31,76 @@ const (
...
 	tierUnknownToMain  = "unknown-to-main"
 	BackendOverrideKey = "CABOOSE_BACKEND_OVERRIDE"
+	CabooseJwtIssuer   = "caboose-client"
 )
 
+// authenticateReq adds authentication to a request when a JWT_SECRET is present as an environment variable.
+func authenticateReq(req *http.Request) (*http.Request, error) {
+
+	// Check for existense of an auth secret.
+	jwtKey := []byte(os.Getenv("JWT_SECRET"))
```
willscott, line 41: we should probably pull this from caboose.Config, rather than only pulling it in as an env var

tieredhashing/tiered_hashing.go
```diff
@@ -241,31 +253,43 @@ func (t *TieredHashing) GetPerf() map[string]*NodePerf {
 	return t.nodes
 }
 
-func (t *TieredHashing) AddOrchestratorNodes(nodes []string) (added, alreadyRemoved, removedAndAddedBack int) {
+func (t *TieredHashing) GetSentinelCid(ip string) (string, error) {
+	if node, ok := t.nodes[ip]; ok {
+		return node.SentinelCid, nil
```
willscott, line 258: should we delete the sentinel once we get it, rather than potentially making the same sentinel request multiple times?

AmeanAsad: I don't see a significant risk of requesting the same sentinel cid multiple times. Although, I think this is dependent on the frequency that we update the sentinel cids. Currently, it is hourly so I think it is enough time to rotate the cids without making too many duplicate requests.

AmeanAsad: One issue I think exists with the current implementation though is that we are only making sentinel cid requests to the top performing nodes in the caboose pool. This is because the sentinel cid request code is triggered when a saturn is picked to fetch a cid. 

So maybe an improvement over this is to have the sentinel cid code instead randomly choose a node from the pool and then delete its sentinel cid? That way we make sure we're making our sentinel cid requests uniformly across all nodes. Thoughts?

pool.go
AmeanAsad: Makes sense, can change that.

fetcher.go
```diff
@@ -48,6 +51,16 @@ var (
 func (p *pool) doFetch(ctx context.Context, from string, c cid.Cid, attempt int) (b blocks.Block, rm tieredhashing.ResponseMetrics, e error) {
 	reqUrl := fmt.Sprintf(saturnReqTmpl, c)
 
+	rand, _ := rand.Int(rand.Reader, big.NewInt(sentinelCidPeriod))
```
aarshkshah1992, line 54: This should be in fetch Resource as that is called for both blocks and CARs.

pool.go
```diff
@@ -28,26 +31,70 @@ const (
 	tierMainToUnknown  = "main-to-unknown"
 	tierUnknownToMain  = "unknown-to-main"
 	BackendOverrideKey = "CABOOSE_BACKEND_OVERRIDE"
+	CabooseJwtIssuer   = "caboose-client"
 )
 
+// authenticateReq adds authentication to a request when a JWT_SECRET is present as an environment variable.
+func authenticateReq(req *http.Request, key string) (*http.Request, error) {
```
aarshkshah1992, line 38: @AmeanAsad We need a unit test for this function too.  Give this function a test http req object and ensure it comes back with what we expect it to.

fetcher.go
willscott: rather than spawning a go-routine / adding a separate percentage / code path for these sentinals, can we merge this with the mirroring of requests and subsequent validation that happens in `checkPool`?

caboose.go
```diff
@@ -101,6 +108,8 @@ const DefaultFetchKeyCoolDownDuration = 1 * time.Minute // how long will a sane
 // however, only upto a certain max number of cool-offs.
 const DefaultSaturnNodeCoolOff = 5 * time.Minute
 
+const DefaultSentinelCidPeriod = int64(200)
```
willscott, line 111: comment needed for what this is, because it isn't really a `Period` - this means that sentinel will be used once every 200 requests, on average, is how i think i'm interrpeting this?

AmeanAsad: Your interpretation is correct, added a comment to clarify the value.

willscott (approved):

PR #122, opened 2023-07-24
forward trace id to saturn

guanzo:
Forwards trace id to saturn if exists


fetcher.go
```diff
@@ -242,6 +244,9 @@ func (p *pool) fetchResource(ctx context.Context, from string, resource string,
 	}
 
 	req.Header.Add("Accept", mime)
+	if traceId != "" {
+		req.Header.Add("X-Trace-Id", traceId)
```
willscott, line 248: you're happy with this as the header?

willscott (approved):

fetcher.go
guanzo: I switched away from `traceparent` because it has a specified format, and I want the new header to be generic instead of specific to bifrost. Do you have any suggestions?

willscott: i think this is fine if it's what you want to adopt on the L1 side.

Is there a particular reason to resist the existing traceparent standard / naming?

guanzo: The `traceparent` value has a specified format that the nginx request id doesn't match: https://www.w3.org/TR/trace-context/#version-format. I think it'd be weird if we used a standard name but not the standard value format. 

That being said it might not be a big deal and in that case I would prefer to keep the `traceparent` naming. wdyt?

willscott: what is the format of the nginx request id?

willscott: (e.g. is it possible to use that as the 'trace-id' portion of a valid `traceparent` header?

guanzo: nginx is 32 hex chars: `444535f9378a3dfa1b8604bc9e05a303`

the `trace-id` is also 32 hex chars which is great. we'd need another random 16 hex chars for the `parent-id`, the other 2 components `version-format` and `trace-flags` can be hardcoded

willscott: we can make a span in caboose and generate a valid `parent-id` to go with the `trace-id` for each sub-request

willscott: that'll also mean that we can differentiate the different 'attempts' that are made within a single bifrost-gateway request context, since they'll have different `parent-id`'s but the same `trace-id`.

guanzo: excellent suggestion, i'll update the PR

guanzo: some issues with this approach:

1. We want to correlate end client requests too, so the `traceparent` header would need to be added to the bifrost-gateway response.
2. `traceparent`s are typically generated by the first caller (arc) and passed down the chain. What we'd be doing instead is generating it at the 2nd caller (technically 3rd if you treat the bifrost lb as a separate caller), then passing it back up and also down the chain. A bit weird but idk if that actually breaks the standard.

guanzo: another thing we can do is have the client add `traceparent` as a query param instead of a header. that would avoid the CORS preflight issue altogether

fetcher.go
```diff
@@ -464,3 +474,18 @@ func getCacheStatus(isCacheHit bool) string {
...
 }
+
+func createTraceparentFromTraceID(traceID string) string {
+	version := "00"
+	parentSpanID := generateRandomHex(8)
+	traceFlags := "01"
+
+	traceparent := fmt.Sprintf("%s-%s-%s-%s", version, traceID, parentSpanID, traceFlags)
+	return traceparent
+}
```
willscott, line 485: instead of doing this manually, can we record this as a span?

* get in a context as you currently do (I would do this at the `Fetch` method in caboose.go)
  * use the parentSpanID from the `requestcontext`
  * turn it into a `spanContext` with [NewSpanContext](https://pkg.go.dev/go.opentelemetry.io/otel/trace#NewSpanContext)
  * Put that span context on the config used throughout caboose using [ContextWithRemoteSpanContext](https://pkg.go.dev/go.opentelemetry.io/otel/trace#ContextWithRemoteSpanContext)

Now you have a span context from upstream on the config for the overall fetch. on each subattempt that caboose makes in this `fetchResource` method you'll indicate that's a sub-span, using:
```
var span trace.Span
ctx, span = tracer.Start(ctx, "attempt")
defer span.End()
```

to get the current sub-span attached in to the request header of the http sub requests you'll do [this](https://github.com/oinume/go-http-client-trace-sample/blob/main/examples/otelhttp_client_trace/main.go#L42-L44)
```
clientTrace := otelhttptrace.NewClientTrace(ctx)
ctx = httptrace.WithClientTrace(ctx, clientTrace)
req, err := http.NewRequestWithContext(ctx, "GET", url, nil)
```

guanzo: The id from `requestcontext` is supposed to be the traceID, no? Since its 32 hex chars. spanID is 16 hex chars. 

I still need a spanID to create the span context. I was hoping to get it from [this span](https://github.com/filecoin-saturn/caboose/blob/cb81e714c3eba70118509f2ef3b0180adb41ef8d/caboose.go#L262), but its spanID is all zeros 🤔 .

willscott: * correct - `requestcontext` will give you the traceID.
* maybe just try https://pkg.go.dev/go.opentelemetry.io/otel/trace#SpanFromContext and get the span id from that - if there is one set in bifrost, you'll inherit it, and otherwise you'll be a child of an all 0 span, which should be okay

PR #123, opened 2023-07-25
parallelise mirrored reqs

aarshkshah1992:
We've never see a significant rate of mirrored requests. Both the Caboose pool tuning and Amean's work now depends on this. This PR parallelises the sending of mirrored requests.

### DIFF
--- a/caboose.go
+++ b/caboose.go
@@ -89,7 +89,7 @@ const DefaultSaturnCarRequestTimeout = 30 * time.Minute
 const defaultMaxRetries = 3
 
 // default percentage of requests to mirror for tracking how nodes perform unless overridden by MirrorFraction
-const defaultMirrorFraction = 0.01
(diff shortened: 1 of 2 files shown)


caboose.go
```diff
@@ -89,7 +89,7 @@ const DefaultSaturnCarRequestTimeout = 30 * time.Minute
 const defaultMaxRetries = 3
 
 // default percentage of requests to mirror for tracking how nodes perform unless overridden by MirrorFraction
-const defaultMirrorFraction = 0.01
```
willscott, line 92: lets leave this at 0.01 for now?

we should isolate how much of the impact is due to the concurrency limitation in place currently rather than changing multiple things at once

aarshkshah1992: @willscott Yeah, makes sense. Done.

willscott (approved):

PR #124, opened 2023-07-26
Dont drain response on errors

aarshkshah1992:
For large responses, could be a problem.

### DIFF
--- a/fetcher.go
+++ b/fetcher.go
@@ -346,10 +346,8 @@ func (p *pool) fetchResource(ctx context.Context, from string, resource string, 
 	wrapped := TrackingReader{resp.Body, time.Time{}, 0}
 	err = cb(resource, &wrapped)
 	received = wrapped.len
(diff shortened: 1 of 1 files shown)


fetcher.go
```diff
@@ -346,10 +346,8 @@ func (p *pool) fetchResource(ctx context.Context, from string, resource string,
 	wrapped := TrackingReader{resp.Body, time.Time{}, 0}
 	err = cb(resource, &wrapped)
 	received = wrapped.len
-	// drain body so it can be re-used.
-	_, _ = io.Copy(io.Discard, resp.Body)
-
 	if err != nil {
+		resp.Body.Close()
```
willscott, line 350: this will double-close. i think we just don't have this line

aarshkshah1992: Fixed

willscott (approved):

PR #129, opened 2023-07-28
Log orchestrator response in debug mode

aarshkshah1992:


willscott (approved):

pool.go
```diff
@@ -251,13 +252,8 @@ func (p *pool) checkPool() {
 				}
 
 				if rand.Cmp(big.NewInt(0)) == 0 {
-					err := p.fetchComplianceCid(node)
-					if err != nil {
```
aarshkshah1992, line 255: This will also be an error as compliance cids wont pass mirror validator

PR #130, opened 2023-08-01
Relative correctness policy in Caboose

aarshkshah1992:
We don't want to evict nodes if their correctness is similar to other nodes even if their absolute correctness is below a certain random threshold as software bugs can cause absolute correctness to dip which will result in us evicting legit nodes.

### DIFF
--- a/tieredhashing/tiered_hashing.go
+++ b/tieredhashing/tiered_hashing.go
@@ -31,8 +31,8 @@ const (
 	correctnessWindowSize = 1000
 
 	// ------------------ CORRECTNESS -------------------
-	// minimum correctness pct expected from a node over a rolling window over a certain number of observations
-	minAcceptableCorrectnessPct = float64(70)
+	// we will evict a node if it's correctness difference relative to other nodes is greater than this threshold
+	correctnessThreshold = float64(25)
 
 	// helps shield nodes against bursty failures
 	failureDebounce = 2 * time.Second
@@ -83,6 +83,9 @@ type TieredHashing struct {
 
 	StartAt  time.Time
 	initDone bool
+
+	OverAllCorrectnessDigest  *rolling.PointPolicy
+	NOverAllCorrectnessDigest float64
 }
 
 func New(opts ...Option) *TieredHashing {
@@ -91,7 +94,7 @@ func New(opts ...Option) *TieredHashing {
 		FailureDebounce:       failureDebounce,
 		LatencyWindowSize:     latencyWindowSize,
 		CorrectnessWindowSize: correctnessWindowSize,
-		CorrectnessPct:        minAcceptableCorrectnessPct,
+		CorrectnessThreshold:  correctnessThreshold,
 		MaxMainTierSize:       maxMainTierSize,
 		NoRemove:              false,
 	}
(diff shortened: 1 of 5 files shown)


tieredhashing/tiered_hashing.go
```diff
@@ -418,10 +423,23 @@ func (t *TieredHashing) isCorrectnessPolicyEligible(perf *NodePerf) (float64, bo
 		return result
 	})
 
+	averageSuccess := t.OverAllCorrectnessDigest.Reduce(func(w rolling.Window) float64 {
```
willscott, line 426: is there any way to memo-ize this value / this function?

doing the reduction over 1000 data points on every single request when the general 'overall correctness' ought to be fairly stable  is a lot of duplicated work.

we can try in staging but i am worried this will be a cpu bottleneck when run at scale as it's written

tieredhashing/tiered_hashing.go
```diff
@@ -421,7 +443,7 @@ func (t *TieredHashing) isCorrectnessPolicyEligible(perf *NodePerf) (float64, bo
 	// should satisfy a certain minimum percentage
 	pct := totalSuccess / perf.NCorrectnessDigest * 100
 
-	return pct, pct >= t.cfg.CorrectnessPct
+	return pct, (t.AverageCorrectnessPct - pct) < t.cfg.CorrectnessThreshold
```
willscott, line 446: is this backwards? ( lets put in a comment of what the boolean means)

if average correctness is 80%
this node correctness is 20%

0.8 - 0.2 = 0.6
0.6 is more than the 20% threshold.

aarshkshah1992: Faceplam. You are right.

aarshkshah1992: @willscott It is correct actually. This boolean is true when we do NOT want to remove a node from the pool i.e. true means node is eligible for the correctness policy.

aarshkshah1992: In your example, bool will be false and so node will be kicked out.

tieredhashing/tiered_hashing.go
```diff
@@ -27,7 +27,7 @@ const (
 	reasonCorrectness = "correctness"
 
 	// use rolling windows for latency and correctness calculations
-	latencyWindowSize     = 100
+	latencyWindowSize     = 50
```
willscott, line 30: this seems unrelated to the larger correctness change? can we do this as a different PR / test them independently?

tieredhashing/tiered_hashing.go
```diff
@@ -31,12 +31,16 @@ const (
...
+	correctnessThreshold = float64(25)
 
 	// helps shield nodes against bursty failures
 	failureDebounce = 2 * time.Second
-	removalDuration = 3 * time.Hour
+	removalDuration = 6 * time.Hour
+)
+
+var (
+	goLogger = golog.Logger("caboose-pool")
```
willscott, line 43: caboose-tiered-hashing ?

tieredhashing/tiered_hashing.go
willscott: can you put in the comment explaining this?

willscott (approved):

aarshkshah1992:
@willscott Whenever you are ready.

aarshkshah1992:
@willscott Addressed your last couple of comments.

PR #131, opened 2023-08-01
Decouple sending compliance cids from sending mirrored traffic for main nodes

aarshkshah1992:
No reason to send compliance cids only when we're mirroring traffic to a main set node.

### DIFF
--- a/pool.go
+++ b/pool.go
@@ -232,22 +232,15 @@ func (p *pool) checkPool() {
 			go func(msg mirroredPoolRequest) {
 				defer func() { <-sem }()
 
-				// see if it is to a main-tier node - if so find appropriate test node to test against.
 				p.lk.RLock()
-				if p.th.NodeTier(msg.node) != tieredhashing.TierMain {
-					p.lk.RUnlock()
-					return
-				}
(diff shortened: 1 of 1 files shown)


pool.go
```diff
@@ -258,6 +251,17 @@ func (p *pool) checkPool() {
 					complianceCidCallsTotalMetric.WithLabelValues("success").Add(1)
 				}
 
+				// see if it is to a main-tier node - if so find appropriate test node to test against.
+				// --- Mirroring
+				p.lk.RLock()
+				if p.th.NodeTier(msg.node) != tieredhashing.TierMain {
```
willscott, line 257: should this be `p.th.NodeTier(node)` ?

aarshkshah1992: @willscott No, what we're checking here is that the original request that caused this mirroring was sent to a main node.

`node` is the node that we will send the mirrored traffic/compliance cid to.

AmeanAsad: I thought `node` refers to the test node we are finding to test the main tier node against and `msg.node` is the main tier node.

AmeanAsad (approved):

PR #132, opened 2023-08-01
update to use otel libraries

willscott:
a friendly update on trace-ID to per latest comments


willscott:
cc @aarshkshah1992 - @guanzo  is out this week, so if you have a moment to review lets see if we can get this over the line sooner.

guanzo (approved):

PR #135, opened 2023-08-02
Increase mirrored requests and prevent main set from starvation during DDOS

aarshkshah1992:
image
Need to investigate high error rate in mirrored requests. Have added logging here for that.

### DIFF
--- a/pool.go
+++ b/pool.go
@@ -264,6 +264,7 @@ func (p *pool) checkPool() {
 
 				cancel()
 				if err != nil {
+					goLogger.Warnw("mirrored request failed", "err", err.Error())
 					mirroredTrafficTotalMetric.WithLabelValues("error").Inc()
 				} else {
 					mirroredTrafficTotalMetric.WithLabelValues("no-error").Inc()
--- a/pool.go
+++ b/pool.go
@@ -259,7 +259,7 @@ func (p *pool) checkPool() {
 					return
 				}
 				p.lk.RUnlock()
-				trialTimeout, cancel := context.WithTimeout(context.Background(), 30*time.Second)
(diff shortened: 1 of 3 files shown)


pool.go
```diff
@@ -259,11 +259,12 @@ func (p *pool) checkPool() {
 					return
 				}
 				p.lk.RUnlock()
-				trialTimeout, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+				trialTimeout, cancel := context.WithTimeout(context.Background(), DefaultSaturnMirrorRequestTimeout)
```
aarshkshah1992, line 262: Tracked down high mirrored request failure rate to the really low timeout of 30 seconds which wont work for fetching entire CARs.  We should use atleast 1/3rd of the actual timeout for CAR requests (now that we make mirrored requests in parallel, this should be fine.)

tieredhashing/tiered_hashing.go
```diff
@@ -428,6 +429,18 @@ func (t *TieredHashing) UpdateMainTierWithTopN() (mainToUnknown, unknownToMain i
 		}
 	}
 
+	// if we don't have enough nodes for the main pool, compromise on number of observations required and fill it up
+	if t.mainSet.Size() < t.cfg.MaxMainTierSize {
```
willscott, line 433: shouldn't this be a `MinMainTierSize` threshold. otherwise it's improperly named and we're essentially forcing an exact singular `MainTierSize`?

aarshkshah1992: @willscott A subsequent PR will be to make this pool unbounded as we've discussed so all performant nodes can join it. Okay if we deal with the naming then ?

willscott (approved):

PR #139, opened 2023-08-04
[WIP] Update go car to handle rootless CARs

aarshkshah1992:


fetcher.go
```diff
@@ -343,6 +345,11 @@ func (p *pool) fetchResource(ctx context.Context, from string, resource string,
 	err = cb(resource, &wrapped)
 	received = wrapped.len
 	if err != nil {
+		// TODO Agree with Bifrost on what this looks like
```
aarshkshah1992, line 348: @willscott @aschmahmann Need to agree on this contract now.

PR #151, opened 2023-08-24
Improve retry logic

aarshkshah1992:
Improve retry logic

### DIFF
--- a/caboose.go
+++ b/caboose.go
@@ -85,11 +85,12 @@ const DefaultLoggingInterval = 5 * time.Second
 const DefaultSaturnOrchestratorRequestTimeout = 30 * time.Second
 
 const DefaultSaturnBlockRequestTimeout = 19 * time.Second
-const DefaultSaturnCarRequestTimeout = 30 * time.Minute
+
+const DefaultSaturnCarRequestTimeout = 1 * time.Minute
 const DefaultSaturnMirrorRequestTimeout = 30 * time.Second
 
 // default retries before failure unless overridden by MaxRetrievalAttempts
-const defaultMaxRetries = 3
+const defaultMaxRetries = 2
 
 // default percentage of requests to mirror for tracking how nodes perform unless overridden by MirrorFraction
 const defaultMirrorFraction = 0.01
--- a/caboose.go
+++ b/caboose.go
@@ -86,7 +86,7 @@ const DefaultSaturnOrchestratorRequestTimeout = 30 * time.Second
 
 const DefaultSaturnBlockRequestTimeout = 19 * time.Second
 
-const DefaultSaturnCarRequestTimeout = 1 * time.Minute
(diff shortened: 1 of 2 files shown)


caboose.go
```diff
@@ -85,11 +85,12 @@ const DefaultLoggingInterval = 5 * time.Second
 const DefaultSaturnOrchestratorRequestTimeout = 30 * time.Second
 
 const DefaultSaturnBlockRequestTimeout = 19 * time.Second
-const DefaultSaturnCarRequestTimeout = 30 * time.Minute
+
+const DefaultSaturnCarRequestTimeout = 1 * time.Minute
```
aarshkshah1992, line 89: Max P95 for a successful CAR fetch over the last 24 hours in prod is 30 seconds:

https://protocollabs.grafana.net/d/bxJz6bJ4k/bifrost-gw-metrics?orgId=1&from=now-24h&to=now&editPanel=134

caboose.go
```diff
@@ -85,11 +85,12 @@ const DefaultLoggingInterval = 5 * time.Second
...
 const DefaultSaturnMirrorRequestTimeout = 30 * time.Second
 
 // default retries before failure unless overridden by MaxRetrievalAttempts
-const defaultMaxRetries = 3
+const defaultMaxRetries = 2
```
aarshkshah1992, line 93: ~99% of ALL requests that succeed do so in the first two retries:

https://protocollabs.grafana.net/d/6g0_YjBVk/bifrost-caboose-staging?orgId=1&from=1692841797171&to=1692863397171&viewPanel=98

fetcher.go
```diff
@@ -236,9 +236,9 @@ func (p *pool) fetchResource(ctx context.Context, from string, resource string,
 	// which is the amount of time without any NEW data from the server, but
 	// that can be added later. We need both because a slow trickle of data
 	// could take a large amount of time.
-	requestTimeout := DefaultSaturnCarRequestTimeout
+	requestTimeout := time.Duration(DefaultSaturnCarRequestTimeout.Milliseconds()/int64(attempt+1)) * time.Millisecond
```
aarshkshah1992, line 239: Reduces timeout duration with each retry.

caboose.go
willscott: does this mean that a large movie won't work anymore?


I don't think this is an acceptable limit.

caboose.go
willscott: this seems fine

fetcher.go
willscott: this seems okay

caboose.go
aarshkshah1992: Hmmm, I see what you mean.

aarshkshah1992: @willscott Reverted this change.

willscott (approved):

PR #154, opened 2023-09-04
Removed unused metrics

aarshkshah1992:
Removes metrics we no longer need as the backpressure PR has landed in main and some of those metrics were put in place to catch old bugs that no longer exist.

### DIFF
--- a/fetcher.go
+++ b/fetcher.go
@@ -309,6 +309,19 @@ func (p *pool) fetchResource(ctx context.Context, from *Node, resource string, m
 	// request life-cycle metrics
 	saturnCallsSuccessTotalMetric.WithLabelValues(resourceType, getCacheStatus(isCacheHit)).Add(1)
 
+	fetchRequestSuccessTimeTraceMetric.WithLabelValues(resourceType, getCacheStatus(isCacheHit), "tcp_connection").Observe(float64(result.TCPConnection.Milliseconds()))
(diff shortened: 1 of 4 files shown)


fetcher.go
```diff
@@ -324,92 +307,20 @@ func (p *pool) fetchResource(ctx context.Context, from *Node, resource string, m
 
 	// trace-metrics
 	// request life-cycle metrics
-
-	fetchRequestSuccessTimeTraceMetric.WithLabelValues(resourceType, getCacheStatus(isCacheHit), "tcp_connection").Observe(float64(result.TCPConnection.Milliseconds()))
```
willscott, line 328: do we not want this l1 breakdown timing anymore?

aarshkshah1992: @willscott It hasn't helped us with anything so far. We know where most time is spent on for hits/misses and we already have these same metrics on Saturn as well. 
I think the e2e tracing work will be more helpful here.

willscott: we want to be able to subtract it from overall request latency to separate out the L1 / caboose proportion from the lassie proportion in the long tail, right?

aarshkshah1992: Hmm, okay, no harm in keeping it then. I've put it back.

aarshkshah1992:
@willscott Can I merge ?

willscott:
waiting for visibility from @AmeanAsad  when he wakes up, since it merge-conflicts with his in-progress PR to the same branch

PR #158, opened 2023-09-11
convert l1 timing headers into otel spans

willscott:
should allow for validation in tracing ux


guanzo:
Could we add a "cacheHit" attribute to the span so we can differentiate if the server timings are for this request or a previous request?

willscott:
done

guanzo (approved):

PR #159, opened 2023-09-13
Add a core attribute to tracing span

willscott:
the configured orchestrator url will also need to be updated in infra, but this will take a 'core' attribute when present on a node and propagate it to the submitted trace for relevant sub requests


willscott:
ref: https://github.com/filecoin-saturn/roadmap/issues/16

DiegoRBaquero (approved):

PR #162, opened 2023-09-19
Test cache affinity

aarshkshah1992:


pool_dynamics_test.go
```diff
@@ -257,46 +252,38 @@ func TestPoolAffinity(t *testing.T) {
...
 
-		// Introduce new nodes by sendng same stats to those nodes.
-		for i := 0; i < poolRefreshNo/2; i++ {
+		orig := make(map[string]string)
+
+		for _, c := range cidList {
+			aff := ch.Caboose.GetAffinity(ctx)
+			if aff == "" {
+				aff = fmt.Sprintf(blockPathPattern, c)
+			}
```
willscott, line 261: this test should only be doing one of these, so shouldn't need this complexity?

willscott (commented):
You've changed this test to make the 'good' nodes better, and then confirm that they're still the same ones for a set of CIDs.

I think this no longer exercises the original point of the test - In the modified version there is no change in the active ring from what I can tell. The goal of this test, i think is to aim to demonstrate that even as some new nodes join the ring, there is relative stability of the ring space otherwise.

aarshkshah1992:
@willscott The goal is to demonstrate that we always get the same nodes for the same cid (as long as the hash ring does not change). Adding nodes that are as good as the original ones will change the hash ring and re reroute content. 

What kind of test do you have in mind here ?

AmeanAsad:
Hey @aarshkshah1992 @willscott I think Aarsh's test here might still serve as a good base case test for cache affinity separate from the original intended test I wrote. 

That way we have two test cases:
- Testing that the hash ring only maps cids to good nodes. 
- Testing stability in the ring space as new good nodes join the pool .

PR #163, opened 2023-09-19
Test cache affinity

aarshkshah1992:
Supersedes https://github.com/filecoin-saturn/caboose/pull/162.

### DIFF
--- a/pool_dynamics_test.go
+++ b/pool_dynamics_test.go
@@ -292,7 +292,7 @@ func TestPoolAffinity(t *testing.T) {
 		}
 
 		// no more than 5 cids from the cid list of 20 should get re-routed (25%)
-		assert.LessOrEqual(t, rerouteCount, 4)
+		assert.LessOrEqual(t, rerouteCount, 5)
 	})
 }
 
--- a/pool_dynamics_test.go
+++ b/pool_dynamics_test.go
@@ -222,7 +222,9 @@ func TestPoolAffinity(t *testing.T) {
 	cidList := generateRandomCIDs(20)
 
 	t.Run("selected nodes remain consistent for same cid reqs", func(t *testing.T) {
-		ch, controlGroup := getHarnessAndControlGroup(t, nodesSize, nodesSize/2)
+		// 80 nodes will be in the good pool. 20 will be added later with the same stats
+		// so 20% of the nodes in the pool will eventually be "new nodes" that will be added later.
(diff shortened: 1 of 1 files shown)


pool_dynamics_test.go
```diff
@@ -265,38 +267,32 @@ func TestPoolAffinity(t *testing.T) {
...
+			for _, n := range badNodes {
+				n := n
+				if n.URL == nodes[0].URL {
+					rerouteCount++
+				}
+			}
 		}
 
+		// no more than 5 cids from the cid list of 20 should get re-routed (25%)
+		assert.LessOrEqual(t, rerouteCount, 4)
```
willscott: should this be 5, rather than 4?

pool_dynamics_test.go
```diff
@@ -265,38 +267,32 @@ func TestPoolAffinity(t *testing.T) {
...
 			}
 			nodes, _ := ch.CabooseActiveNodes.GetNodes(aff, ch.Config.MaxRetrievalAttempts)
 
-			// We expect that the candidate nodes are part of the "good nodes" list.
-			assert.Contains(t, goodNodes, nodes[0])
+			for _, n := range badNodes {
+				n := n
+				if n.URL == nodes[0].URL {
+					rerouteCount++
+				}
```
willscott, line 290: should 'reroute' be incremented when this doesn't match?

AmeanAsad: I believe reroute is meant to keep track of how often caboose returns a new joining node as the candidate node to fetch a cid in which case caboose is "rerouting" where it fetches that cid from. In that case, the increment here is good imo.

aarshkshah1992: Yeah what Amean said.

aarshkshah1992: We're iterating over bad nodes here.

AmeanAsad:
Hey @aarshkshah1992 I think we should change notation of "bad" and "good" nodes in the affinity tests to maybe just "baseNodes" and "newNodes". I know I started with that notation in the test but I think its better to change it to make the test a bit more clear.

aarshkshah1992:
@AmeanAsad Done.

pool_dynamics_test.go
aarshkshah1992: Done.

AmeanAsad (approved):
Looks good!

